PR #78, opened 2019-02-06
Fix QCM graphic media display

VincentCATILLON:
## Detailed purpose of the PR

This PR fixes the Graphic QCM media not displayed.

### Result and observation

kapture 2019-02-06 at 12 38 39

- [ ] **Breaking changes ?** 
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing


src/modules/uri.test.js
```diff
@@ -0,0 +1,22 @@
...
+    expect(result).toEqual(expected);
+  });
+
+  it('should handle http uri', () => {
+    const result = getCleanUri('http://domain.tld');
+    expect(result).toEqual(expected);
+  });
+
+  it('should handle https uri', () => {
+    const result = getCleanUri('http://domain.tld');
```
chrisdugne: https

src/modules/uri.js
```diff
@@ -0,0 +1,6 @@
+// @flow strict
+
+/* eslint-disable import/prefer-default-export */
+
+export const getCleanUri = (originalUri: string): string =>
+  originalUri.replace(/(http:|https:|)\/\//g, 'https://');
```
chrisdugne: the point to use `//` is to automatically get `http` or `https` prefixes depending on your domain, not to have CORS issues. 
I'm not sure it's  good idea to override and use `https` in every case....

chrisdugne:
```
Now that SSL is encouraged for everyone and doesn’t have performance concerns, this technique is now an anti-pattern. If the asset you need is available on SSL, then always use the https:// asset.
```
https://www.paulirish.com/2010/the-protocol-relative-url/

ok

VincentCATILLON: Yes but we don't have any domain in mobile app.

By default, http calls are blocked by iOS, look at the official [doc](https://facebook.github.io/react-native/docs/network):

capture d ecran 2019-02-06 a 13 28 41

src/modules/uri.test.js
VincentCATILLON: done

PR #94, opened 2019-02-14
Split tests in few files

VincentCATILLON:
## Detailed purpose of the PR

This PR splits tests and fix all e2e tests broken by plugging store.

### Result and observation

Everything should work right now. 
Thus, this PR changes these points:

1. last question was ending on a correction screen without life
2. correction screen (in error), was ending in level end screen in success
3. adds adaptive course in fixtures for e2e tests

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing


VincentCATILLON:
Missing thing, the adaptive fixtures should not work :) @chrisdugne

PR #95, opened 2019-02-14
Fix brew cache linked

VincentCATILLON:
## Detailed purpose of the PR

This PR fixes the way we are using brew dependencies from cache (link so) and install them by default.

### Result and observation

End 2 end tests should be green and 4min faster.

- [ ] **Breaking changes ?**  
- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [ ] Unit testing


.circleci/config.yml
```diff
@@ -82,8 +82,14 @@ jobs:
       - run:
           name: Brew dependencies
           command: |
-            brew tap wix/brew
-            brew install applesimutils
+            if brew ls --versions applesimutils > /dev/null; then
```
VincentCATILLON: If applesimutils is restored by cache, we'll only link to save time (2min50 -> 0min02)

PR #101, opened 2019-02-19
Fix header spacing bug on android

Fausto95:
## Detailed purpose of the PR

clue-header-spacing

Fixes the header spacing  in the clue screen on Android

### Result and observation

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


Fausto95:
[(10) ETQU, je vois le clue si il existe](https://trello.com/c/DpUhENsp/1050-10-etqu-je-vois-le-clue-si-il-existe)

src/containers/card-flippable.js
```diff
@@ -9,7 +9,8 @@ import {BrandThemeContext} from '../components/brand-theme-provider';
 const HEIGHT = 270;
 const styles = StyleSheet.create({
   container: {
-    flex: 1
+    flex: 1,
+    padding: theme.spacing.base
```
VincentCATILLON: Don't understand what removing padding fixes something related to centered text ?

VincentCATILLON: It this for the cards animation overflow which was truncated when rotating ?

Fausto95: This is not about the text being centered or not, it's about putting padding where it shouldn't be, because if we let the padding stay there, the view that holds the text will be pushed to somewhere

Fausto95:
FYI: the only step that failed was the `buil_ios` but I cannot retry it from circle

PR #102, opened 2019-02-19
Add video subtitles

VincentCATILLON:
## Detailed purpose of the PR

This PR adds video subtitles support.

**iOS**:

kapture 2019-02-20 at 16 39 04

**Android**:

kapture 2019-02-20 at 16 43 46

### Result and observation

You should be able to see subtitles, disable or enable it, in Android and iOS.

**NB**: Only in e2e mode until the API is ready.

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


VincentCATILLON:
Missing things:
- unit tests
- e2e tests
- manage bad uri for subtitles

src/__fixtures__/lessons.js
```diff
@@ -15,6 +15,7 @@ export const lessonWithVideo: Lesson = {
   mimeType: 'application/vimeo',
   ref: 'les_E1TniqOCB',
   type: 'video',
+  subtitleRef: 'sub_dacb7432-9ce1-4711-95fc-7963c599c9cb',
```
Volubyl: it might be usefull to track the differences btw the actual bundled discipline returned value (by the API i mean) and the updated fixture.

It won't be easy for @esa-coorp or @godu  to understand what prop we've  added/updated/deleted

VincentCATILLON: This property is already defined in API and used by the Mooc:
https://github.com/CoorpAcademy/coorpacademy/blob/f4656048b302a099740b54d1803142ecc580366d/core/public/js/services/KollectiveService.js#L124

PR #108, opened 2019-02-20
Splitting resources lecture into a component

Fausto95:
## Detailed purpose of the PR

This PR splits the resources preview into a component so we can re-use it anywhere

### Result and observation

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


Fausto95:
I will need this in Context too

Fausto95:
@chrisdugne 's version only consider rendering a video or a pdf and never both at the same time.
SO we'll merge this until he fixes it on what is currently working on!

VincentCATILLON:
Closing it because it's outdated (seen with @chrisdugne )

chrisdugne:
deprecated, see https://github.com/CoorpAcademy/mobile/pull/111

PR #115, opened 2019-02-25
Adding course card on Homescreen

alanlanglois:
## Detailed purpose of the PR

Adding cards for courses on the Home screen

### Result and observation
card

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


src/components/catalog-item.js
```diff
@@ -0,0 +1,261 @@
...
+        testID={`background-image-${testID}`}
+        source={image}
+        style={[styles.background, {minHeight: currentStyle.minHeight}]}
+      >
+        <LinearGradient
+          testID={`gradient-${testID}`}
+          colors={['rgba(0,0,0,0)', 'rgba(0,0,0,0)', 'rgba(0,0,0,0.9)']}
+          style={[styles.content, {padding: currentStyle.paddingGradient}]}
+        >
+          {badgeLabel && (
```
Fausto95: All of these checks makes this component unreadable!
Could you please turn all of them into small components inside this file with a beautiful name so if some we come here to refactor we know what's going on?

src/components/catalog.js
```diff
@@ -14,27 +21,73 @@ type Props = {|
...
   props: Props;
 
   handlePress = (item: Item) => () => this.props.onPress(item);
 
+  isChapter = (item: Item): boolean => !item.hasOwnProperty('modules');
+
+  getImageURI(item: Item): string {
+    let image;
+    if (this.isChapter(item)) {
+      image = item.poster && item.poster.mediaUrl;
```
Fausto95: returning the images URI inside each check isn't it way better than creating a `let image`variable ?

src/components/catalog-item.js
```diff
@@ -0,0 +1,261 @@
...
+  isInfinite,
+  authorType,
+  isCertified,
+  displayMode,
+  testID
+}: Props) => {
+  const mode: DisplayMode = displayMode ? displayMode : DISPLAY_MODE.COVER;
+
+  const cardStyle: GenericStyleProp = {
+    titleSize: 16,
```
Fausto95: cardStyle & coverStyle have only one dynamic property, mind  putting them in `styles` variable and then 
`const cardStyle = { ...styles.cardStyle, paddingGradient: theme.spacing.small};`
`const coverStyle = { ...styles.cardStyle, paddingGradient: theme.spacing.base};`
 ?

src/components/catalog.js
alanlanglois: It would return undefined in some case, and we don't want it

Fausto95: `return item.poster && item.poster.mediaUrl || ''` would work in this case?
If the left hand operation evaluates to null or undefined we gonna return an empty string

src/components/catalog-item.js
Fausto95: Yeah, my bad, I made a typo, what I really meant was:
`const cardStyle = { ...styles.cardStyle, paddingGradient: theme.spacing.small};`
`const coverStyle = { ...styles.coverStyle, paddingGradient: theme.spacing.base};`

Fausto95:
Yes, I understand the downside, but at least the code is more readable now, and maybe we can use these new components later on, who know?!

PR #116, opened 2019-02-25
Cards fetch at start

VincentCATILLON:
## Detailed purpose of the PR

This PR adds logic to fetch cards at start:
- Cards service
- Cards data layer
- Cards fixtures (for runtime and e2e)
- Cards Redux state

### Result and observation

You should see on home the previous courses (no chapter yet).

- [x] **Breaking changes ?** Catalog props and onboarding/bescherelle fixtures path
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing


src/layer/data/cards.js
```diff
@@ -0,0 +1,43 @@
...
+import bescherelle from '../../__fixtures__/__temporary__/bescherelle-course';
+import type {SupportedLanguage} from '../../translations/_types';
+import type {Discipline, DisciplineCard, ChapterCard, Cards} from './_types';
+
+export const fetchCards = (
+  language: SupportedLanguage
+): Promise<Cards<DisciplineCard | ChapterCard>> => {
+  let disciplinesObject: {[key: string]: Discipline} = {};
+
+  if (__E2E__) {
```
godu: `__E2E__` n'est jamais vrai. Jest mock ne semble pas marcher comme tu le penses.

VincentCATILLON: Done, as it was imported before the mock definition, it was always false

PR #121, opened 2019-02-26
Adding Catalog-Item component and card refacto

alanlanglois:
## Detailed purpose of the PR
Adding catolog-item component splitted in 4 components:
 - catalog-item-badge
 - catalog-item-footer
 - catalog-item-author
 - image-gradient

Refactoring card component splitting it to card and card-header component 
and rewriting card-scalable container to have it working in accordance to this refactoring.

### Result and observation
- [ ] **Breaking changes ?**  
      card-scalable container now need to add  the header and card component

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


src/components/card.js
```diff
@@ -1,103 +1,39 @@
...
-    <View style={styles.content}>
-      {children}
-      <Gradient height={theme.spacing.large} color={theme.colors.white} />
-    </View>
-  </View>
-);
+const Card = ({children, style, testID, isDeckCard}: Props) => {
+  if (isDeckCard) {
+    return (
+      <View style={[style, styles.container, Platform.OS === 'ios' && {overflow: 'hidden'}]}>
```
VincentCATILLON: You can do this thing in styles like this:
```
container: {
    ...Platform.select({
      ios: {
        overflow: 'hidden'
      }
    }),
}
```

src/components/card.js
```diff
@@ -1,103 +1,39 @@
...
+        <View style={[styles.container]} testID={testID}>
```
VincentCATILLON: Remove array please

VincentCATILLON: Can you please check what is the aim of this `View` ? IMO it's redundant with his parent

src/components/card.js
```diff
@@ -1,103 +1,39 @@
...
+          {children}
+        </View>
+      </View>
+    );
+  } else {
+    return (
+      <View style={[style, styles.container]} testID={testID}>
+        <View style={[styles.container, {overflow: 'hidden'}]}>{children}</View>
```
VincentCATILLON: Same for this.
Don't you do the overflow hidden only on Platform.OS === 'ios' ? (please refer to the comment above relating `Platform.select()`)

src/components/catalog-item-author.js
```diff
@@ -0,0 +1,53 @@
...
+  author: {
+    color: theme.colors.white,
+    fontSize: 12,
+    textShadowColor: 'rgba(0, 0, 0, 1)',
+    textShadowRadius: 2
+  }
+});
+
+const CatalogItemAuthor = ({authorType, fontSize, testID}: Props) => {
+  if (authorType === 'coorp') {
```
VincentCATILLON: Can use constant please ?

src/components/catalog-item-author.js
```diff
@@ -0,0 +1,53 @@
...
+    return (
+      <View style={styles.authorContainer}>
```
VincentCATILLON: COMMENT: Can be container

src/components/catalog-item-author.js
```diff
@@ -0,0 +1,53 @@
...
+        <Text testID={`author-${testID}`} style={[styles.author, {fontSize: fontSize}]}>
+          COORP <Text style={{fontWeight: theme.fontWeight.bold}}>ORIGINAL</Text>
```
VincentCATILLON: Can you please add a style for this in styles object?

src/components/catalog-item-author.js
```diff
@@ -0,0 +1,53 @@
...
+    textShadowColor: 'rgba(0, 0, 0, 1)',
+    textShadowRadius: 2
+  }
+});
+
+const CatalogItemAuthor = ({authorType, fontSize, testID}: Props) => {
+  if (authorType === 'coorp') {
+    return (
+      <View style={styles.authorContainer}>
+        <Text testID={`author-${testID}`} style={[styles.author, {fontSize: fontSize}]}>
```
VincentCATILLON: ES6: When the constant is the same than the property, you can do this: `{fontSize}`

src/components/catalog-item-author.js
```diff
@@ -0,0 +1,53 @@
...
+          COORP <Text style={{fontWeight: theme.fontWeight.bold}}>ORIGINAL</Text>
+        </Text>
+      </View>
+    );
+  } else {
+    return (
+      <View style={styles.authorContainer}>
+        <Text
+          testID={`author-${testID}`}
+          style={[styles.author, {fontSize: fontSize, fontWeight: theme.fontWeight.bold}]}
```
VincentCATILLON: Same as above for ES6, and style for fontWeight please

src/components/catalog-item-badge.js
```diff
@@ -0,0 +1,64 @@
...
+import theme from '../modules/theme';
+import type {Chapter, Discipline} from '../layer/data/_types';
+import Text from './text';
+import {BrandThemeContext} from './brand-theme-provider';
+
+export type Item = Discipline | Chapter;
+
+type Props = {|
+  label: string,
+  testID: string,
```
VincentCATILLON: testID is mandatory ?

src/components/catalog-item-badge.js
```diff
@@ -0,0 +1,64 @@
...
+    flexDirection: 'row',
+    backgroundColor: theme.colors.white,
+    overflow: 'hidden',
+    borderBottomEndRadius: theme.radius.medium,
+    height: 19
+  },
+  badge: {
+    padding: theme.spacing.micro,
+    paddingLeft: theme.spacing.tiny,
+    paddingRight: theme.spacing.tiny + theme.spacing.micro,
```
VincentCATILLON: ? why doing this ? if we have a theme for spacing, we should follow it :)

src/components/catalog-item-badge.js
```diff
@@ -0,0 +1,64 @@
...
+      <BrandThemeContext.Consumer>
+        {brandTheme => (
+          <Text
+            testID={testID}
+            style={[
+              styles.badge,
+              {
+                minWidth: minWidth,
+                minHeight: minHeight,
+                fontSize: fontSize,
```
VincentCATILLON: Same as above (ES6)

src/components/catalog-item-footer.js
```diff
@@ -0,0 +1,115 @@
...
+};
+
+type Props = {|
+  title: string,
+  subtitle: string,
+  progression?: Progression,
+  isAdaptive: boolean,
+  isCertified?: boolean,
+  testID: string,
+  style?: Style
```
VincentCATILLON: Can you please define 3 separate props ? `style` prop is usually a `GenericStyleProp` (refer to `./flow-typed/_custom-defs.js`)

src/components/catalog-item-footer.js
```diff
@@ -0,0 +1,115 @@
...
+              testID={`certified-${testID}`}
+              color={theme.colors.white}
+              height={style.subtitleSize}
+              width={style.subtitleSize}
+            />
+          </View>
+        )}
+      </View>
+      {progression && (
+        <View style={styles.progressionBar} testID={`progressBar-${testID}`}>
```
VincentCATILLON: COMMENT: Its usually in kebab case, but don't take it in consideration

src/components/catalog-item.js
```diff
@@ -0,0 +1,121 @@
...
+  onPress,
+  title,
+  subtitle,
+  progression,
+  image,
+  badge,
+  isAdaptive,
+  authorType,
+  isCertified,
+  displayMode,
```
VincentCATILLON: Can be defined there: `displayMode = DISPLAY_MODE.COVER`

src/components/catalog-item.js
```diff
@@ -0,0 +1,121 @@
...
+  testID
+}: Props) => {
+  const mode: DisplayMode = displayMode ? displayMode : DISPLAY_MODE.COVER;
```
VincentCATILLON: (See comment above, you can remove it)

src/components/catalog-item.js
```diff
@@ -0,0 +1,121 @@
...
+    titleSize: 22,
+    subtitleSize: 16,
+    progressBarHeight: 2,
+    authorSize: 12,
+    paddingGradient: theme.spacing.base,
+    badgeW: 45,
+    badgeH: 20,
+    badgeFontSize: 11,
+    minHeight: 265
+  };
```
VincentCATILLON: Can you please define these 2 styles in a `styles` object and use it with reference (`styles.card`) ? You can so remove the switch below

src/components/catalog-item.js
```diff
@@ -0,0 +1,121 @@
...
+        image={image}
+        style={{padding: currentStyle.paddingGradient}}
+        minHeight={currentStyle.minHeight}
+      >
+        {badgeLabel && (
+          <Badge
+            label={badgeLabel}
+            minWidth={currentStyle.minWidth}
+            minHeight={currentStyle.minHeight}
+            fontSize={currentStyle.badgeFontSize}
```
VincentCATILLON: IMO these 3 values should be given with `style` prop

src/components/catalog-item.js
```diff
@@ -0,0 +1,121 @@
...
+            testID={`badge-${testID}`}
+          />
+        )}
+        {authorType && (
+          <CatalogItemAuthor
+            authorType={authorType}
+            fontSize={currentStyle.authorSize}
```
VincentCATILLON: Same for this

src/components/catalog-item.js
```diff
@@ -0,0 +1,121 @@
...
+
+  const badgeLabel =
+    badge && badge !== '' ? badge.charAt(0).toUpperCase() + badge.slice(1) : undefined;
+
+  return (
+    <TouchableHighlight testID={testID} onPress={onPress}>
+      <ImageGradient
+        testID={testID}
+        image={image}
+        style={{padding: currentStyle.paddingGradient}}
```
VincentCATILLON: If you are defining a style in `styles` object, you can directly refer to it: `style={styles.myStyle}`

src/components/catalog-item.js
```diff
@@ -0,0 +1,121 @@
...
+        minHeight={currentStyle.minHeight}
```
VincentCATILLON: (And so, minHeight can be defined in `style` too)

src/const.js
```diff
@@ -14,6 +14,13 @@ export const RESOURCE_TYPE: {
   IMG: 'img'
 };
 
+export const DISPLAY_MODE: {
```
VincentCATILLON: COMMENT: I'd prefer something more explicit: `CARD_DISPLAY_MODE`

src/types.js
```diff
@@ -34,3 +36,8 @@ export type GenericContent = $Exact<{|
   ...GenericContentBase,
   type: ContentType
 |}>;
+
+export type Progression = {|
+  current: number,
+  count: number
+|};
```
VincentCATILLON: This type is already defined in `ProgressionBar` component :)

src/components/card.js
alanlanglois: The style differ depending the card is displayed in the deck-swipe or not, it fixes the following  bug we have on master on android : 
image

src/components/card.js
alanlanglois: It's the only way I found to get the shadow working on both platform, shadow beeing applied to the parent through `style`.

src/types.js
alanlanglois: Is it? I just see two props "current" and "count"

src/components/card.js
alanlanglois: Not in this case, the conditional overflow hidden on iOS is just there to fix the bug mention above.

src/components/card.js
VincentCATILLON: Yes, my purpose was that you can create a style with `Platform.select()`. This will make the same as u did

src/components/card.js
VincentCATILLON: So can you create a dedicated style for this ? styles.hidden for example

src/types.js
VincentCATILLON: You can reuse the props, but it's ok :)

src/components/card.js
VincentCATILLON: Ok but create a style for this please, ie: `styles.overflowHidden`

VincentCATILLON:
Nice work dude

VincentCATILLON:
(Please fix unit tests: `yarn test:unit -u` ;) )

PR #122, opened 2019-02-26
Rework card's data-layer type

godu:


VincentCATILLON:
Merging locally to fix lint issues.

PR #124, opened 2019-02-27
Use the fresh new VTT endpoint (subtitles)

VincentCATILLON:
## Detailed purpose of the PR

This PR adds the new VTT endpoint in production to display subtitles without any mock.

Not to be confused with this kind of VTT:

### Result and observation

You can see subtitles in video lessons (depending on video data).
On e2e environment, you can see them in videos in basic/adaptive course.

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


src/__fixtures__/discipline-bundle/adaptive.js
```diff
@@ -16,7 +16,7 @@ const lessons = [
   createVideo({
     ref: 'les_1',
     description: 'First video',
-    subtitleRef: 'sub_dacb7432-9ce1-4711-95fc-7963c599c9cb'
+    subtitleRef: 'ref_subtitle_01'
```
Fausto95: sub_subtitle_01 change it everywhere

VincentCATILLON: No no, this is the name of the real file in API

Fausto95: This is useless I know, but to respect the same pattern with other projects because subtitles's refs start with `sub_`

VincentCATILLON: Ok but it's not my deal, it's API name file, i won't change it if it's named `toto.vtt`, it's data

PR #126, opened 2019-02-27
Fetch dachboard's data from MOOC

godu:
You can try to modify favorites playlist with this user : https://mobile-staging.coorpacademy.com/analytics/users/activity/user-5c751aa68f2713001b30e1b3


src/layer/data/cards.js
```diff
@@ -1,40 +1,74 @@
...
+const LIMIT_CARDS = 5;
 
+export const fetchCards = async (
+  language: SupportedLanguage,
+  host: string,
+  token: string
+): Promise<Cards> => {
   if (__E2E__) {
-    disciplinesObject = {
+    let disciplinesObject: {[key: string]: Discipline} = {
```
VincentCATILLON: should be const as its not updated earlier

src/layer/data/cards.js
```diff
@@ -1,40 +1,74 @@
...
-    disciplinesObject = {
-      ...onboarding.disciplines,
-      ...bescherelle.disciplines
-    };
-  }
 
-  // Implement fetching logic here
-  const disciplines = Object.keys(disciplinesObject).map(key => disciplinesObject[key]);
-  const cards = createDisciplinesCards(disciplines);
+    // Implement fetching logic here
```
VincentCATILLON: Comment can be removed as you did the fetching logic

src/layer/data/cards.js
```diff
@@ -1,40 +1,74 @@
...
+  );
+  const result: {hits?: Cards} = await response.json();
+
+  return result.hits || [];
+};
+
+const LIMIT_CARDS = 5;
 
+export const fetchCards = async (
+  language: SupportedLanguage,
```
VincentCATILLON: 👍

src/redux/middlewares/cards.js
```diff
@@ -20,6 +20,10 @@ import type {Cards, Level, Chapter} from '../../layer/data/_types';
 
 type State = StoreState;
 
+const TOKEN =
+  'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyIjoiNWM3NTFhYTY4ZjI3MTMwMDFiMzBlMWIzIiwiYnJhbmQiOiJtb2JpbGUiLCJob3N0IjoiaHR0cHM6Ly9tb2JpbGUtc3RhZ2luZy5jb29ycGFjYWRlbXkuY29tIiwiaWF0IjoxNTUxMTc4ODEwLCJleHAiOjIwMDAwMDAwMDAsImF1ZCI6Im1vYmlsZSIsImlzcyI6ImNvb3JwYWNhZGVteS1qd3QifQ.MRRlxll4lX5lel45679Nj-4IDGsVRwybxsUxbrl20Qs';
+const HOST = 'https://mobile-staging.coorpacademy.com';
```
VincentCATILLON: This one is already defined in `brand-theme-provider` and should be used or mutualized (through redux?)

godu: In the future, we'll need to store in redux the `(token, brand)` tuple. As you said, the card's middleware need to get host and token from store.
But i don't think it's a good idea to work on it before the authentification feature.

PR #132, opened 2019-03-04
Question Draggable Feature

Volubyl:
## Detailed purpose of the PR

Add question draggable question type

### Result and observation

draggable-question

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [X] Manual testing
- [X] Unit testing
- [X] E2e testing


__e2e__/questions.test.js
```diff
@@ -8,6 +8,20 @@ const rightAnswer = async () => {
   await element(by.id('button-validate')).tap();
 };
 
+const selectQCMRIghtDragItem = async () => {
+  await element(by.id(`unSelected-choice-play_store`)).tap();
```
Fausto95: unselect is an existing word in en, so: `unselect` instead of unSelect

package.json
```diff
@@ -55,7 +55,7 @@
     "@coorpacademy/nova-icons": "^2.6.0",
     "@coorpacademy/player-services": "^2.4.6",
     "@coorpacademy/player-store": "^2.5.8",
-    "@coorpacademy/progression-engine": "^10.3.6",
```
Fausto95: ??

src/components/drop-zone.js
```diff
@@ -0,0 +1,79 @@
...
+    borderColor: theme.colors.gray.light,
+    backgroundColor: theme.colors.gray.extra,
+    borderRadius: theme.radius.common,
+    marginBottom: theme.spacing.tiny
+  },
+  emptyContent: {
+    justifyContent: 'center',
+    alignContent: 'center',
+    alignItems: 'center',
+    height: 60
```
VincentCATILLON: Please use padding

src/components/drop-zone.js
```diff
@@ -0,0 +1,79 @@
...
+        testID={`selected-choice-${item.value || index}`}
+        onPress={this.handlePress(item)}
+      >
+        {item.label}
+      </QuestionChoice>
+    ));
+
+    const hasNoSelectedChoices = mappedSortedChoices.length === 0;
+
+    const content = hasNoSelectedChoices ? (
```
VincentCATILLON: I think it's easier to read like that:

```
<View style={[styles.dropZone, hasNoSelectedChoices && styles.emptyContent]}>
  {hasNoSelectedChoices && (...)}
  {!hasNoSelectedChoices && (...)}
</View>
```

src/components/drop-zone.stories.js
```diff
@@ -0,0 +1,37 @@
...
+  describe('DropZone', () => {
+    it('should handle onItemPress callback', () => {
+      const handleItemPress = jest.fn();
+      const component = renderer.create(<DropZone choices={choices} onPress={handleItemPress} />);
+      const questionItem = component.root.find(
+        el => el.props.testID === `selected-choice-play_store`
+      );
+      questionItem.props.onPress();
+      expect(handleItemPress.mock.calls.length).toBe(1);
+      expect(handleItemPress.mock.calls[0]).toEqual([choices[0]]);
```
VincentCATILLON: Nice one !

src/const.js
```diff
@@ -92,3 +94,10 @@ export const ENGINE: {
   LEARNER: 'learner',
   MICROLEARNING: 'microlearning'
 };
+
+export const FONT_SIZE: {
+  [string]: FontSize
+} = {
+  SMALL: 'small',
+  REGULAR: 'regular'
+};
```
VincentCATILLON: IMO these values should be in theme with values instead of string, ie:

```
{
  ...
  fontSize: {
    small: 13,
    regular: 15,
    large: 17
  }
}
```

src/types.js
```diff
@@ -23,6 +17,7 @@ export type MimeType =
 export type SpaceType = 'micro' | 'tiny' | 'small' | 'base' | 'large' | 'xlarge';
 
 export type CardType = 'tip' | 'keyPoint' | 'correction';
+export type FontSize = 'regular' | 'small';
```
VincentCATILLON: As comment above

__e2e__/questions.test.js
```diff
@@ -8,6 +8,20 @@ const rightAnswer = async () => {
   await element(by.id('button-validate')).tap();
 };
 
+const selectQCMRIghtDragItem = async () => {
+  await element(by.id(`unselected-choice-play_store`)).tap();
```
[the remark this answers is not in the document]
Volubyl: a compromise would be to use the _id as testID unique identifier.

src/const.js
VincentCATILLON: These values should definitly in theme (`src/modules/theme`)

PR #134, opened 2019-03-05
Dashboard / catalog

alanlanglois:
## Detailed purpose of the PR
Adding courses/disciplines dashboard to the home screen

### Result and observation
  dashboard

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


yarn.lock
```diff
@@ -1487,14 +1487,6 @@
     fuzzy-matching "^0.4.3"
     lodash "^4.17.4"
 
-"@coorpacademy/progression-engine@^10.3.6":
```
VincentCATILLON: 👍

src/components/catalog-item-author.js
```diff
@@ -27,23 +29,34 @@ const styles = StyleSheet.create({
...
+        </View>
+      );
+    case AUTHOR_TYPE.CUSTOM:
+      return (
+        <View style={styles.container}>
+          <Text testID={`author-${testID}`} style={[styles.author, style]}>
+            {authorName && authorName.toUpperCase()} CREATION
+          </Text>
+        </View>
+      );
```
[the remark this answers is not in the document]
alanlanglois: Good point.
For the 2 cents question: 'CREATION' is not in bold on zeplin https://app.zeplin.io/project/5c6d8091c49d450d828bb204/screen/5c7d646686d8b8629225fdbb

src/layer/data/_types.js
```diff
@@ -126,7 +126,7 @@ export type CardSkill = {|
 export type CardAuthor = {|
   ref: string,
   label: string,
-  authorType: string
+  authorType: 'coorp' | 'verified' | 'custom' | 'marketplace'
```
VincentCATILLON: Could be AuthorType so, but it's ok :)

alanlanglois: Yes but I couldn't find any other types imported from types.js, I thought it was maybe because we don't want dependency between the two types files.
Next time I'll remplace it by AuthorType though :)

PR #138, opened 2019-03-06
Question slider

Fausto95:
This PR brins support slider question type

capture d ecran 2019-03-06 a 11 07 36

question-slider

question-slider

- [x] **Breaking changes ?**  
       Added new props to `src/screens/Question.js`, `components/QuestionChoices.js` & `components/Questions.js`

- [x] **Extra dependency ?**
         - https://github.com/CoorpAcademy/react-native-slider (This is a Slider component);
         - https://react-native-training.github.io/react-native-elements;

I've just installed `react-native-elements` juste to use their `slider`component and I know its annoying but we have no choices because all of the availables `slider` components  on the are just animated views, they kinda bug when we re-render de component multiple times with random values(this is a normal behavior of `@coorpacademy/player-store`  when it deals with questions).

Another solution was to use the built-in react-native's `slider`component, this one works really well, but as we do not have full control of it when it comes to styling so this is the reason why I've considered not to use it (FYI: the thumb bar is is so small by default and unfortunately  we cannot change it 🙃).

So react-native-elements's `slider` components fits well in our use cases, it does not bug and its gives us a testID out-of-box that way we can easily to test(e2e) the component.

You might be wondering, why should we install a whole library just to use one specific component? I know this increases the size of our bundle with all these unused component, unfortunately we have no solutions yet. 
AFAIK, tree-shaking might be landing this year, at least I hope so!

https://github.com/facebook/metro/issues/227

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


Fausto95:
[ETQU je peux répondre à une question de type slider](https://trello.com/c/mSpdgDvE/1123-etqu-je-peux-r%C3%A9pondre-%C3%A0-une-question-de-type-slider)

PR #140, opened 2019-03-06
Safearea props on Screen component

alanlanglois:
## Detailed purpose of the PR
Adding the ability to remove SafeArea, can be useful for some use case such as the dashboard. 

### Result and observation
**before:**

OR

**now:**

Adding the prop noSafeArea when the screen needs to be fullscreen:
```
<Screen testID="home-screen" noSafeArea>
   <Home onCardPress={this.handleCardPress} />
</Screen>
```

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


src/components/screen.js
```diff
@@ -38,25 +39,46 @@ class Screen extends React.PureComponent<Props> {
...
+                showsHorizontalScrollIndicator={false}
+                testID={testID}
+                ref={this.handleRef}
+              >
+                {children}
+              </ScrollView>
+            )}
+          </SafeAreaView>
         )}
-      </SafeAreaView>
```
VincentCATILLON: You can make something more simple :

const CustomView = noSaferArea ? View : SafeAreaView;

[...]

<CustomView /> 

instead of duplicating blocks

alanlanglois: 🙏

VincentCATILLON:
Good job dude 🤝

PR #141, opened 2019-03-07
Using a pretty slider

Fausto95:
Using a pretty slider instead of the built-in

# Before:

capture d ecran 2019-03-07 a 10 12 43

# After 

capture d ecran 2019-03-07 a 10 52 30

- [ ] **Breaking changes ?**  

- [x] **Extra dependency ?**
      `@coorpacademy/react-native-slider`

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


src/components/question-slider.js
```diff
@@ -49,6 +50,25 @@ const styles = StyleSheet.create({
...
+    width: 30,
+    height: 30,
+    borderRadius: 30 / 2,
+    // shadowColor
+    shadowOffset: {
+      width: 3,
+      height: 3
+    },
+    shadowOpacity: 0.4,
+    // shadowRadius
```
VincentCATILLON: you can use STYLE spreaded there imported from box component :) and so remove all shadow properties

src/containers/with-slider.js
```diff
@@ -34,14 +34,22 @@ const withSlider = <P>(
   class ComponentWithSlider extends React.Component<Props, State> {
     props: Props;
 
-    state: State = {
-      value: this.props.value,
-      minValue: this.props.minValue,
-      maxValue: this.props.maxValue
-    };
+    constructor(props: Props) {
```
VincentCATILLON: Type your state please

src/containers/with-slider.js
```diff
@@ -34,14 +34,22 @@ const withSlider = <P>(
...
+    }
+
+    componentDidMount() {
+      const {value, minValue, maxValue} = this.props;
+      this.updateState(value, minValue, maxValue);
+    }
 
     componentWillUnmount() {
-      this.updateState(0, 0, 0);
+      this.updateState(0, 0, 10);
```
VincentCATILLON: **comment**: maybe using a key on slider will avoid state usage

PR #143, opened 2019-03-07
bump store to fix default resources playing

chrisdugne:
## Detailed purpose of the PR

actions on play() a resource where not dispatched properly, because no default resource was selected

capture d ecran 2019-03-07 a 14 39 23

### Result and observation

the fix is bumped within `player-store`
https://github.com/CoorpAcademy/components/commit/a29843dc7e2e13078f754fc09f367d6a7846d4d6#diff-cba28ccf83284a6721764fbbc71c6881R54

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


chrisdugne:
[bug play resource: cannot play default video/pdf](https://trello.com/c/O3uZxUk3/1142-bug-play-resource-cannot-play-default-video-pdf)

PR #147, opened 2019-03-08
Fetch bundle

godu:
Fetch `discipline-bundle` from MOOC's API for each cards.


src/redux/middlewares/discipline-bundle.js
```diff
@@ -7,24 +7,34 @@ import type {StoreState} from '../store';
...
 
 const createMiddleware = ({services}: Options): Middleware<State, Action, Dispatch<Action>> => ({
-  dispatch
+  dispatch,
+  getState
 }: MiddlewareAPI<State, Action, Dispatch<Action>>) => (
   next: Dispatch<Action>
 ): Dispatch<Action> => (action: Action) => {
   if (action.type === FETCH_REQUEST) {
+    const token = getToken(getState());
```
Fausto95: const state = getState();

src/layer/data/core.js
```diff
@@ -98,40 +97,50 @@ export const storeDisciplineBundle = async (
...
   }
 
-  if (Object.keys(bescherelle.disciplines).includes(ref)) {
-    return Promise.resolve(bescherelle);
-  }
-
-  // Implement fetching logic here
-  return Promise.reject(new Error('API fetching not supported yet.'));
+  const response = await fetch(
+    `${host}/api/v2/disciplines/bundle?lang=${language}&conditions={"ref": ["${ref}"]}`,
```
Fausto95: `cross-fetch` does not support `qs: {}` so we could
```js
 qs: {
   lang: language,
   conditions: {
     ref: [ ...]
   }
}
```

Fausto95:
Provide some descriptions @godu 🤨

src/redux/middlewares/discipline-bundle.js
VincentCATILLON: I agree, to avoid two calls to getState()

src/layer/data/core.js
godu: nope... https://fetch.spec.whatwg.org/#fetch-api
I'll try to find a url polyfill.

VincentCATILLON: `Accept-Language` HTTP header is not enough ? https://developer.mozilla.org/fr/docs/Web/HTTP/Headers/Accept-Language

src/layer/data/core.js
```diff
@@ -98,40 +97,50 @@ export const storeDisciplineBundle = async (
...
+      return Promise.resolve(disciplinesBundle);
+    }
+
+    if (Object.keys(onboarding.disciplines).includes(ref)) {
+      return onboarding;
+    }
+
+    if (Object.keys(bescherelle.disciplines).includes(ref)) {
+      return bescherelle;
+    }
```
VincentCATILLON: These 2 courses should be removed (and __temporary__ folder too) as they will be fetched

godu: Si sauf quand tu es connecté. La langue de l'utilisateur en base prend le dessus.

src/layer/data/core.js
godu: Yes, but user's language setting overwrite this value.

godu: header => user's language => query

PR #148, opened 2019-03-08
Fetch five unique cards

godu:
If wishlist and recommendation return the same ref, the dashboard display cards less then five cards.


src/layer/data/cards.js
```diff
@@ -51,12 +58,15 @@ export const fetchCards = async (
 
     return cards;
   }
-  const [favorites, recommendations] = await Promise.all([
-    fetchFavoriteCards(token, host, language),
-    fetchRecommendationCards(token, host, language)
-  ]);
 
-  return [...favorites, ...recommendations].slice(0, LIMIT_CARDS);
+  const favoritesP = fetchFavoriteCards(token, host, language);
```
```js
const favorites = await fetchFavoriteCards
Const recommendations = await fetchRecommendationCards
```

src/layer/data/cards.js
```diff
@@ -36,6 +36,13 @@ const fetchRecommendationCards = async (
   return result.hits || [];
 };
 
+const uniqBy = <O: {+[string]: mixed}, P: $Keys<O>>(key: P, array: Array<O>): Array<O> =>
```
VincentCATILLON: This should be on a helper file as its generic and not directly related to cards

src/layer/data/cards.js
godu: Nope, here, I want to start fetches and wait recommendations just if favorites doesn't have at least 5 items.
Recommendations call is the longer one, if we can skip this time, we win.

PR #154, opened 2019-03-11
Sort by completion

godu:
Select the right level when we tap on discipline card
Sort card by completion.

https://trello.com/c/c6s5CLej/1111-etqu-jai-un-dashboard-qui-me-permet-dacc%C3%A9der-aux-cours


src/utils/index.test.js
```diff
@@ -0,0 +1,20 @@
+// @flow strict
+
+import {uniqBy} from '.';
+
+describe('Utils', () => {
+  it('uniqBy', () => {
+    expect(uniqBy(v => v.toString(), [1, 2, 2])).toEqual([1, 2]);
```
Fausto95: Why do I think that the expected result should be `['1', '2']`;
`mapper: (obj: O) => string` -> `expect(uniqBy(v => v.toString(), [1, 2, 2])).toEqual(['1', '2']);`

godu: I don't want to recast values. 
```js
uniqBy<T>(..., Array<T>): Array<T>
```

godu: Like a kind of filter

PR #164, opened 2019-03-15
support some tags

Volubyl:
## Detailed purpose of the PR

Support those tags emedded in the content :

- `br` 
- `u`
-`i` 
- ...

Currently we do not support the `<sup>`,`<sub>` tags

### Result and observation

Look we have underlined and italic text 

Screenshot 2019-03-15 at 15 57 10

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [X] Manual testing
- [ ] Unit testing


src/components/html.js
```diff
@@ -45,6 +57,33 @@ const Html = ({
     h6: {fontSize},
     img: imageStyle
   };
+
+  const renders = {
```
Fausto95: `renderers`

src/components/html.js
```diff
@@ -45,6 +57,33 @@ const Html = ({
...
+            color: htmlAttribs.color
+          }}
+        >
+          {componentChildren}
+        </Text>
+      );
+    },
+
+    // eslint-disable-next-line react/display-name
+    br: (htmlAttribs, componentChildren) => {
```
Fausto95: Actually br is supported if you `</br>`

src/components/html.js
Fausto95: change this var name

PR #171, opened 2019-03-18
A Real(Native) Splash Screen

Fausto95:
splash-screen (1) (1)-min

- [ ] Manual testing
- [ ] Unit testing


android/app/src/main/res/values/colors.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <color name="blue">#4481EB</color>
+    <color name="primary_dark">#4481EB</color>
```
VincentCATILLON: are you sure about these 2 values ? should not be `primary` and `primary_dark` ? and so avoid having 2 values ?

VincentCATILLON: this value is never used

src/components/carousel-card.js
```diff
@@ -60,37 +59,28 @@ class CarouselCard extends React.Component<Props> {
...
+          </Html>
+        </View>
+      </View>
     );
   }
 }
 
 export {CarouselCard as Component};
-export default withLayout(CarouselCard);
+export default CarouselCard;
```
VincentCATILLON: These updates are not related to splashscreen.
Btw, 2 exports for the same thing

src/containers/carousel.js
```diff
@@ -100,8 +108,8 @@ class Carousel extends React.PureComponent {
   extractKey = (item: Item) => item.iconName;
 
   render() {
-    const {state: {items}} = this;
-    const {width} = Dimensions.get('window');
+    const {state: {items}, props: {layout}} = this;
+    const width = layout && layout.width;
```
VincentCATILLON: These updates are not related to splashscreen

android/app/src/main/res/values/colors.xml
Fausto95: This is nor related to our code!
This is a react native work around!

PR #175, opened 2019-03-19
Resume progression

godu:
Store lastProgressionId to resume le last created progression from a given content.

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing


src/redux/actions/cards.js
```diff
@@ -124,11 +124,22 @@ export const selectCard = (item: DisciplineCard | ChapterCard): StoreAction<Acti
     switch (item.type) {
       case CARD_TYPE.CHAPTER: {
         try {
+          // Resume progression
+          const lastProgression = await services.Progressions.findLast(
+            'microlearning',
```
VincentCATILLON: use constant ENGINE please

src/redux/actions/cards.js
```diff
@@ -141,11 +152,22 @@ export const selectCard = (item: DisciplineCard | ChapterCard): StoreAction<Acti
           return dispatch(selectCardFailure(item, 'Course has no level'));
         }
         try {
+          // Resume progression
+          const lastProgression = await services.Progressions.findLast(
+            'learner',
```
VincentCATILLON: Same as above (constant)

PR #176, opened 2019-03-19
add storybook case to show the rendering behavior according to HTML tag

Volubyl:
## Detailed purpose of the PR

Just add/update storie to ensure HTML coverage

### Result and observation

A question choice with this kind of content 

```html
 <i> italic input </i> and <s> input </s>
```

with be display like this 

Screenshot 2019-03-19 at 15 47 58

and this kind of content 

```html
  Les deux adverbes ont<s>une terminaison</s> qui se <u>prononce "ament"</u>, mais leur orthographe diffère. On écrit ainsi :
    <br/>– <i>vaillamment</i>, parce que l'adverbe est formé à partir d’un <b>adjectif en <i>-ant</i></b> (<i>vaillant</i>) ;
```

will be display like this 

Screenshot 2019-03-19 at 15 48 26

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [X] Manual testing
- [ ] Unit testing


src/components/html.stories.js
```diff
@@ -9,26 +9,9 @@ storiesOf('Html', module)
   .add('Default', () => (
     <Html fontSize={20}>
```
Volubyl: Just created a test case closer than a real use case

VincentCATILLON: You are not dealing with all tags

Volubyl: this example shows the bahavior of the HTML tags which are required to deal with eg :<b> <i> <u> <s> <br>

see : https://trello.com/c/ulHz7Z5K/1154-mobile-les-balises-html-ne-fonctionnent-pas-partout-o%C3%B9-%C3%A7a-devrait-l%C3%AAtre

PR #183, opened 2019-03-21
Lesson notification

VincentCATILLON:
## Detailed purpose of the PR

This PR adds notification on lesson tab when we have a lesson which is not seen yet.

### Result and observation

Kapture 2019-03-21 at 12 24 45

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing


VincentCATILLON:
End to end are failing, not due to my code. Should fail on master too.

Ping @chrisdugne

PR #193, opened 2019-03-22
Clear on logout

godu:
On logout :
- partial clear of store. We want to preserve permissions and navigation.
- clear all asyncstorage.


src/modules/version.js.flow
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+declare module.exports: {|
+  brandch: string,
```
adamska27: brandch?

.eslintrc
```diff
@@ -19,6 +19,7 @@
     }
   },
   "rules": {
+    "import/prefer-default-export": "off",
```
VincentCATILLON: These rules might be in https://github.com/CoorpAcademy/eslint-plugin-coorpacademy/tree/master/config (react-native file). IMO, we should talk about it after the MVP

__e2e__/utils.js
```diff
@@ -21,16 +21,6 @@ export const reloadApp = async (
   }
 };
 
-export const resetApp = (
```
VincentCATILLON: Why did you remove this ? It's used by e2e to manage permissions

VincentCATILLON: See: https://github.com/wix/Detox/blob/master/docs/APIRef.DeviceObjectAPI.md#2-set-runtime-permissions to have more informations

godu: It's not used anymore.

VincentCATILLON: Oh ok my bad the naming were really similar with `reloadApp` ^^

PR #206, opened 2019-03-25
Fixes the cards order at mount

VincentCATILLON:
## Detailed purpose of the PR

This PR fixes cards order when viewing a resource in correction cards.

### Result and observation

Kapture 2019-03-25 at 15 54 38

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [ ] Unit testing


src/screens/correction.js
```diff
@@ -66,9 +66,11 @@ type Props = {|
 type State = {|
   lives?: number,
   isLoading: boolean,
-  hasLivesBeenAnimated: boolean
+  hasLivesBeenAnimated: boolean,
+  isResourceViewed: boolean
 |};
 
+// @todo move it in class
```
godu: In class ? What do you mean ?

VincentCATILLON: I mean in the component, to avoid dealing with props injected like this

PR #210, opened 2019-03-26
Create function required to display the progression bar

Volubyl:
## Detailed purpose of the PR

To Long to try to understand : 

We update the dashboard card on each progression update and accordingly .

Long drawn version :  

### Result and observation

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [X] Manual testing
- [X] Unit testing


src/__fixtures__/completion.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+import type {Completion} from '../layer/data/_types';
+
+const createCompletion = ({stars, current = 0}: {stars: number, current?: number}): Completion => {
```
chrisdugne: au lieu de `{stars: number, current?: number}` tu peux:
```
type CompletionExtension = $Shape<Completion>;
```

godu: Bien vu

PR #216, opened 2019-03-27
fix keyboard on textinput + lockscreen portrait

adamska27:
https://trello.com/c/aWIIqVFn/1191-bug-ios-etqu-je-veux-voir-ce-que-j%C3%A9cris-dans-un-champ-free-texte

https://trello.com/c/vaiRLcGI/1117-etqu-je-veux-que-l%C3%A9cran-reste-statique-si-je-suis-en-mode-paysage-sauf-pour-lire-une-vid%C3%A9o-en-full-screen

## Detailed purpose of the PR

- fix keyboard hide text input
- lock orientation to portrait on mobile, except for the video in fullscreen

### Result and observation

- [ ] **Breaking changes ?**  

- [x] **Extra dependency ?**
- react-native-device-info
- react-native-keyboard-aware-scroll-view
### Testing Strategy

- [x] Manual testing
- [ ] Unit testing


src/components/question.js
```diff
@@ -87,7 +88,11 @@ const Question = ({
   const mediaUri = media && media.url && getCleanUri(media.url);
 
   return (
-    <View testID="question" style={styles.container}>
+    <KeyboardAwareScrollView
```
VincentCATILLON: Screen has scroll too

PR #220, opened 2019-03-27
Fix version comparing using semver :hammer_and_wrench:

AdrieanKhisbe:
## Detailed purpose of the PR
Fix intempestive autoupdate

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


src/modules/store.js
```diff
@@ -2,7 +2,7 @@
 
 import {Platform} from 'react-native';
 import fetch from 'cross-fetch';
-import compare from 'semver-compare';
+import * as semver from 'semver';
```
godu: Leave my tree shacking alone !

```diff
- import * as semver from 'semver';
+ import {gt} from 'semver';
```

AdrieanKhisbe: yes indeed

PR #224, opened 2019-03-28
Ignore silent mode on iOS when playing a video

VincentCATILLON:
## Detailed purpose of the PR

Currently, the video is in silent mode when the silent mode switch on iOS is enabled.

**Note**: It also fixes the PP used by the app (Adhoc was used in Development mode).

### Result and observation

Keep the switch in enabled state, and run a video. You should listen the 🔉 

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [ ] Unit testing


ios/Coorpacademy.xcodeproj/project.pbxproj
```diff
@@ -293,6 +293,13 @@
 			remoteGlobalIDString = 3D3CD9181DE5FBD800167DC4;
 			remoteInfo = "jschelpers-tvOS";
 		};
+		4017314D224CCF7A003EB148 /* PBXContainerItemProxy */ = {
```
godu: Pourquoi cette diff ?

VincentCATILLON: Dunno, it should be deleted before, but should be handled (as my Xcode says).

It also embeds updates on PP used by the app because we should used dev PP when building in dev mode :)

PR #227, opened 2019-03-28
Catalogitem fix

alanlanglois:
## Detailed purpose of the PR
Force the subtitle to take 2 lines if too long
 
### Result and observation
**before:**

**after:**

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


src/components/catalog-item-footer.js
```diff
@@ -11,7 +11,7 @@ import theme from '../modules/theme';
 import Text from './text';
 import ProgressionBar from './progression-bar';
 
-type Props = {|
+export type Props = {|
```
VincentCATILLON: useless export as it's not used

alanlanglois: yes ;)

PR #240, opened 2019-04-01
Fix tool tip and center picto

Fausto95:
# Screenshots coming sooooonnn

# B4

Capture d’écran 2019-04-01 à 17 35 05

# After

Capture d’écran 2019-04-01 à 17 32 05

- [ ] Manual testing
- [ ] Unit testing


src/components/tooltip.js
```diff
@@ -36,21 +39,32 @@ const styles = StyleSheet.create({
...
+    paddingHorizontal: theme.spacing.base
+  },
   corner: {
     position: 'absolute',
     bottom: 0,
     left: 0,
     width: 16,
-    height: 14
+    height: 14,
+    opacity: 20,
```
alanlanglois: Opacity shouldn't be between 0 and 1?

Fausto95: Ouch, I was playing around with this for test, don't even need it!

PR #252, opened 2019-04-08
Update @coorpacademy/xxx packages

Volubyl:
## Detailed purpose of the PR

Update coorpacademy packages and related dependencies

### Result and observation

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [X] Manual testing
- [ ] Unit testing


package.json
```diff
@@ -117,17 +117,17 @@
     "@storybook/react-native": "^4.0.6",
     "aws-sdk": "^2.355.0",
     "babel-core": "^7.0.0-bridge.0",
-    "babel-eslint": "^9.0.0",
-    "babel-jest": "^23.6.0",
+    "babel-eslint": "^10.0.1",
+    "babel-jest": "^23.4.2",
```
VincentCATILLON: Strange to downgrade...

package.json
```diff
@@ -142,7 +142,8 @@
     "react-native-version": "^2.7.0",
     "react-test-renderer": "16.4.1",
     "regenerator-runtime": "^0.12.1",
-    "shelljs": "^0.8.3"
+    "shelljs": "^0.8.3",
+    "@babel/preset-flow": "^7.0.0"
```
VincentCATILLON: should be with other babel dependencies (added manually 😨 )

Volubyl: right. I also had to update the eslint-xxx packages in order to fit with the new `@coorpacademy/eslint-plugin-coorpacademy` version

PR #146, opened 2019-03-08
Authentication through QR code

VincentCATILLON:
## Detailed purpose of the PR

This PR adds authentication through QR code with permissions handler.

Use this QR code to authenticate:
capture d ecran 2019-03-08 a 17 01 03

(Source: [https://user-images.githubusercontent.com/15608581/53570829-bbcdae80-3b67-11e9-80c9-12775f07422c.png](https://user-images.githubusercontent.com/15608581/53570829-bbcdae80-3b67-11e9-80c9-12775f07422c.png))

### Result and observation

You can authenticate use QR code to authenticate 

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Already covered by tests
- [ ] Manual testing
- [ ] Unit testing

PR #253, opened 2019-04-09
anonymous signIn done✓

Fausto95:
- [ ] Manual testing
- [x] Unit testing


src/redux/actions/authentication.js
```diff
@@ -55,6 +56,18 @@ export const signIn = (token: string): StoreAction<Action> => async dispatch =>
   }
 };
 
+export const signInAnonymous = (): StoreAction<Action> => async dispatch => {
+  const {_bodyInit: token} = await fetch('https://up.coorpacademy.com/api/v1/anonymous/mobile', {
```
Fausto95: It's bizarre that I need to do this `const {_bodyInit: token} = ...` in order to get the response

src/screens/home.js
```diff
@@ -63,7 +63,7 @@ class HomeScreen extends React.PureComponent<Props, State> {
     const action = this.props.onLogoLongPress && (await this.props.onLogoLongPress());
     // When we log out, we reset the entirely stack
     // To avoid router directions and other stuffs
-    if (action) this.props.navigation.reset([NavigationActions.navigate({routeName: 'Splash'})], 0);
```
Fausto95: By the way @VincentCATILLON, IDK why you've refactored this code, but this is not the good way to handle this!
Whenever something goes wrong we're going to stuck in the splash's screen blue dead color

Fausto95: When logging out, we do delete every user's local data, so we are sure that we do not have any stored token, redirecting the user to the splash screen is useless

src/components/authentication.js
```diff
@@ -108,11 +114,22 @@ const Authentication = ({onPress, onAssistancePress}: Props) => (
...
-        <Space type={SPACE.SMALL} />
+        <Space type={SPACE.BASE} />
+        <View style={styles.centeredContent}>
+          <Html
+            fontSize={theme.fontSize.large}
+            style={styles.startDemo}
+            onLinkPress={onStartDemoPress}
+            anchorTextColor={theme.colors.white}
+          >
+            {`No learning platform? <a href="/"><b><u>Start demo</u></b></a>`}
```
Fausto95: Must pull translation from player web

VincentCATILLON: Create a translation key for this

src/components/authentication.js
```diff
@@ -75,6 +76,11 @@ const styles = StyleSheet.create({
     fontSize: theme.fontSize.medium,
     color: theme.colors.white
   },
+  startDemo: {
+    margin: 0,
```
VincentCATILLON: Is this really useful ?

src/redux/actions/authentication.js
```diff
@@ -55,6 +56,15 @@ export const signIn = (token: string): StoreAction<Action> => async dispatch =>
   }
 };
 
+export const signInAnonymous = (): StoreAction<Action> => async dispatch => {
+  const response = await fetch('https://up.coorpacademy.com/api/v1/anonymous/mobile', {
+    method: 'POST'
+  });
+  const token = await response.text();
+  // $FlowFixMe
+  return signIn(token)(dispatch);
```
VincentCATILLON: should not be `return dispatch(signIn(token))` ?

src/screens/authentication.js
```diff
@@ -43,14 +53,19 @@ class AuthenticationScreen extends React.PureComponent<Props> {
     return (
       <Screen testID="authentication-screen" noScroll noSafeArea>
         <StatusBar barStyle="light-content" backgroundColor={TOP_COLOR} />
-        <Authentication onPress={this.handlePress} onAssistancePress={this.handleAssistancePress} />
+        <Authentication
+          onStartDemoPress={this.handleSignInAnonymous}
```
VincentCATILLON: handleStartDemoPress

src/screens/home.js
```diff
@@ -63,7 +63,8 @@ class HomeScreen extends React.PureComponent<Props, State> {
...
+    if (action)
+      this.props.navigation.reset([NavigationActions.navigate({routeName: 'Authentication'})], 0);
```
VincentCATILLON: https://github.com/CoorpAcademy/mobile/pull/146/commits/7d0fda7720c8a7a3c29dd673a7f3c595b9f5cbf2#diff-a77f71d8a99d3eec962380c09d1ae727R53 ;)

Fausto95: Yea, @VincentCATILLON , my bad, squashing commits makes it only appear the pr author as the main man

src/redux/actions/authentication.js
Fausto95: That's a great shortcut, but dont work for tests!

PR #258, opened 2019-04-10
Update confettis animation lib

Fausto95:
- [ ] Manual testing
- [x] Unit testing


src/components/level-end.js
```diff
@@ -280,11 +282,17 @@ class LevelEnd extends React.PureComponent<Props> {
                 </View>
               </View>
             </ScrollView>
-            {isSuccess && (
```
VincentCATILLON: Are you sure about the condition ? Because it was on `isSuccess` before which is not used anymore there

PR #260, opened 2019-04-11
serial sync progressions

Fausto95:
Instead of sync all progressions at once, we sync the one by one!

- [ ] Manual testing
- [ ] Unit testing


flow-typed/p-map.js
```diff
@@ -0,0 +1,7 @@
+declare module 'p-map' {
```
VincentCATILLON: should definitly be in `flow-typed/npm` and installed through `flow-typed` cli, not manually.

btw, the file name should be `p-map_vx.x.x.js` where `x.x.x` can be defined relating to the version you're using

PR #273, opened 2019-04-24
fix jest not defined error when launching storybook

Volubyl:
## Detailed purpose of the PR

The title tells the truth.

### Result and observation

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


src/utils/tests.js
```diff
@@ -29,9 +30,11 @@ export const sleep = (duration: number = 10) =>
...
 export const fakeLayout: Layout = {width: 320, height: 768};
 
-export const createFakeAnalytics = (): AnalyticsState => ({
-  logEvent: jest.fn()
-});
+export const createFakeAnalytics = (): AnalyticsState => {
+  return {
+    logEvent: __TEST__ ? jest.fn() : () => {}
+  };
+};
```
VincentCATILLON: remove return and return directly the object please

PR #274, opened 2019-04-24
Universal links support

VincentCATILLON:
## Detailed purpose of the PR

This PR adds universal links support which lets the user authenticate himself from his browser.

### Result and observation

Install the app, tap on your magic link, it should open the app and authenticate yourself automatically.

**NB**: We'll try to use referrer to authenticate automatically after store installation, but it might need some extra providers like branch, firebase etc, or an API to store the data and match it at the first launch.

To test it, download the version provided by this PR, go to https://dynamiclink.herokuapp.com and tap on `production` to emulate the magic link behavior from your mooc account

- [ ] **Breaking changes ?**
- [x] **Extra dependency ?** app-store-link

### Testing Strategy

- [x] Manual testing
- [ ] Unit testing


VincentCATILLON:
Needs: https://github.com/CoorpAcademy/coorpacademy-lambda/pull/202

PR #275, opened 2019-04-25
Refacto/resource + handle video in question component

Volubyl:
## Detailed purpose of the PR

This PR meant to make the component resource more reusable and to make it a "real entry point" for resource (`LessonType`) handling. Now, if we have to work with a PDF, a video or either an Image we only need to use the `Resource` component to handle the display of such resources.

The also purpose of this PR is to make the `Resource`'s consumer (Context, Lesson, Question-Choice, ...) less coupled  to HOC's like `withLayout`.

#PRODUCT PART 

Screenshot 2019-05-02 at 10 00 37

### REFACTOOOOORING 

BEFORE SITUATION SCHEMA :
Image_d%u2019iOS

AFTER 
IMG_20190415_161743

### Result and observation

No noticeable UI changes

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [X] Manual testing
- [ ] Unit testing


src/components/context.js
```diff
@@ -62,79 +60,83 @@ class Context extends React.PureComponent<Props> {
   props: Props;
 
   renderResource = (resource: Media) => {
-    const {layout} = this.props;
-    const height = layout && layout.width / (16 / 9);
-    const width = layout && layout.width;
     const resourceType = resource && resource.type;
```
Volubyl: this switch is not a real good pattern i know but my first attempt to clean up this was not really efficient so i decied to leave this because it brings more clarity

src/modules/uri.js
```diff
@@ -1,12 +1,21 @@
...
-import type {Lesson} from '@coorpacademy/progression-engine';
+import type {LessonType} from '@coorpacademy/progression-engine';
 import {RESOURCE_TYPE} from '../const';
+import type {MimeType} from '../types';
 
 export const getCleanUri = (originalUri: string): string =>
   originalUri && originalUri.replace(/(http:|https:|)\/\//g, 'https://');
 
-export const getResourceUrl = (resource: Lesson): string | void => {
+export const getResourceUrl = (
```
Volubyl: updated 'cause using a shape type is lighther than using a whole Lesson object when we only need few props

MelkMerle:
@Bertrand: petit retour graphique sur ton screenshot: on dirait que le padding sur les côtes n’est pas le même que pour les propositions de réponses. Ça serait bien d’aligner non ?

__e2e__/lesson.test.js
```diff
@@ -87,33 +89,33 @@ describe('Lesson', () => {
...
+      //   // //   await weExpect(element(by.id('video-fullscreen-shrink'))).toBeVisible();
+      //   // // });
+
+      //   // // it('should shrink the video', async () => {
+      //   // //   await element(by.id('video-fullscreen-shrink')).tap();
+      //   // //   await waitFor(element(by.id('video-container'))).toBeVisible();
+      //   // //   await weExpect(element(by.id('video-container'))).toBeVisible();
+      //   // //   await weExpect(element(by.id('video-container-fullscreen'))).toBeNotVisible();
+      //   // //   await weExpect(element(by.id('video-fullscreen-expand'))).toBeVisible();
+      //   // // });
```
VincentCATILLON: Please remove these duplicate `//`

src/components/question-choice.js
```diff
@@ -113,11 +113,13 @@ const QuestionChoice = ({
...
               <View style={[styles.imageContainer]}>
-                <ImageBackgroundScalable
+                <Resource
                   testID={prefixTestID && `${prefixTestID}${mediaSuffix}`}
-                  source={{uri: mediaUri}}
+                  type={RESOURCE_TYPE.IMG}
+                  url={mediaUri}
                   style={styles.image}
                   resizeMode="contain"
+                  containerStyle={{flex: 1}}
```
VincentCATILLON: use a style please

src/components/question.js
```diff
@@ -91,7 +92,35 @@ const Question = ({
     type === QUESTION_TYPE.TEMPLATE
       ? choices.length === userChoices.filter(choice => choice).length
       : userChoices.length > 0;
-  const mediaUri = media && media.url && getCleanUri(media.url);
+
+  const getMediaInfos = (mediaItem?: Media) => {
```
VincentCATILLON: Return type please

VincentCATILLON: I think this method should be in a helper (`src/modules/medias` ?) and be used by other components like context (and there, obviously)

VincentCATILLON: IMO we should split this function into 3 functions (one per field) to limit responsability and return types

src/components/resource.js
```diff
@@ -1,65 +1,97 @@
...
+import type {WithLayoutProps} from '../containers/with-layout';
 import {RESOURCE_TYPE} from '../const';
 
 import Video from '../containers/video-controlable';
 import {getCleanUri} from '../modules/uri';
+
 import Preview, {EXTRALIFE} from './preview';
 
 type Props = {|
-  type: LessonType | typeof EXTRALIFE,
```
VincentCATILLON: <3

src/components/resource.js
```diff
@@ -1,65 +1,97 @@
...
+      subtitles,
+      layout,
+      testID,
+      thumbnail = '',
+      resizeMode = 'contain',
+      extralifeOverlay = false,
+      style
+    } = this.props;
+
+    const height = layout && layout.width / (16 / 9);
```
VincentCATILLON: The only one place we got 16/9 calculation, i love it bru'

src/components/resource.js
```diff
@@ -1,65 +1,97 @@
...
+      case RESOURCE_TYPE.IMG: {
+        return (
+          <View style={{width: layout && layout.width}}>
+            <ImageBackground
               testID={testID}
+              source={{uri: getCleanUri(url)}}
+              resizeMode={resizeMode}
+              style={{
+                ...style,
+                height: (style && style.height) || height,
```
VincentCATILLON: please add comment

src/containers/video-controlable.js
```diff
@@ -129,7 +130,7 @@ class VideoControlable extends React.PureComponent<Props, State> {
   render() {
     return (
       <React.Fragment>
-        <NavigationEvents onWillBlur={this.handleBlur} />
+        {!__STORYBOOK__ ? <NavigationEvents onWillBlur={this.handleBlur} /> : null}
```
VincentCATILLON: Tricky one but needed for storybook :(

src/screens/correction.js
```diff
@@ -186,6 +186,9 @@ class CorrectionScreen extends React.PureComponent<Props, State> {
         <NavigationEvents onDidFocus={this.handleDidFocus} />
         <StatusBar barStyle="light-content" backgroundColor={backgroundColor} />
         <Correction
+          containerStyle={{
+            flex: 1
+          }}
```
VincentCATILLON: use style please

src/containers/video-controlable.js
VincentCATILLON: add comment about storybook and rn navigation incompatibility please

src/components/context.js
Fausto95: Why isn't it a good pattern?

Volubyl: I meant not in general but here in our cas we could do better to get relevant data to display our resource without using x different render method

PR #281, opened 2019-05-03
Authentication flow

VincentCATILLON:
## Detailed purpose of the PR

This PR integrates a magic link tutorial into a brand new authentication worflow.

### Dependencies

- [x] Merge the [PR](https://github.com/CoorpAcademy/coorpacademy/pull/14071) into the mooc
- [ ] Merge the mobile [PR](https://github.com/CoorpAcademy/mobile/pull/281) with the `pull-translations` script updated
- [ ] Merge the [PR](https://github.com/CoorpAcademy/components/pull/1573) into the components repo (cleanup of legacy translations)

### Result and observation

Kapture 2019-05-07 at 19 27 17

You should make all the stuff defined in AT.

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing (not yet but will be)


src/components/authentication-step.js
```diff
@@ -0,0 +1,55 @@
...
+    color: theme.colors.gray.dark
+  }
+});
+
+const AuthenticationStep = ({icon: Icon, step, description}: Props) => (
+  <View style={styles.container}>
+    <Icon color={BLUE_COORP_LIGHT} width={27} height={27} />
+    <Space type="small" />
+    <Html fontSize={theme.fontSize.small} style={styles.step} isTextCentered>
+      {translations.formatString(
```
Fausto95: create a variable for this pls, doing this in the render method makes it really unreachable and unclear

src/components/authentication.js
```diff
@@ -1,154 +1,134 @@
...
-    height: 21
+  contentFooter: {
+    fontWeight: theme.fontWeight.bold,
+    fontSize: theme.fontSize.large,
+    color: theme.colors.white,
+    textAlign: 'center'
   }
 });
 
 export const TOP_COLOR = BLUE_COORP_DARK;
```
Fausto95: needless

src/components/carousel.js
```diff
@@ -0,0 +1,126 @@
...
+  paginationDotContainer: {
+    marginHorizontal: theme.spacing.tiny / 2,
+    padding: 2,
+    borderColor: theme.colors.white,
+    borderWidth: 1,
+    borderRadius: PAGINATION_DOT_WIDTH
+  }
+});
+
+class Carousel<ItemT> extends React.PureComponent<Props<ItemT>> {
```
Fausto95: since you did `type Props<ItemT>` in your type declaration, why are you doing this multiple times?

src/components/authentication-details.js
```diff
@@ -0,0 +1,139 @@
...
+  },
+  button: {
+    fontWeight: theme.fontWeight.bold,
+    fontSize: theme.fontSize.large,
+    color: BLUE_COORP_LIGHT,
+    textAlign: 'center'
+  }
+});
+
+export const TOP_COLOR = BLUE_COORP_DARK;
```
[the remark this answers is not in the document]
VincentCATILLON: Yes, to avoid having a desynchro between the screen status bar and the component if we'll change the color. We shouldn't assume that the component is using `BLUE_COORP_DARK` color, but we should assume that the top color can change, and so, adjust the status bar to be the same.

src/components/authentication.js
VincentCATILLON: Same comment as above

src/components/carousel.js
VincentCATILLON: ?? Because this is the way to give a type from a class to their props: https://github.com/facebook/react-native/blob/1151c096dab17e5d9a6ac05b61aacecd4305f3db/Libraries/Lists/FlatList.js#L332

Fausto95: Flow's fault!

src/components/authentication-details.js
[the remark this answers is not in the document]
Fausto95: I do not agree!
IMV you're just creating this variables so can figure out how they're applied which is good but needless. 
Anyways, keep it if you wish!

src/modules/theme.js
```diff
@@ -27,8 +27,9 @@ export type FontSizeType =
   | 'medium'
   | 'regular'
   | 'large'
-  | 'extraLarge'
-  | 'xxlarge';
+  | 'xlarge'
```
Volubyl: was it really important to refacto this ?

src/components/authentication.js
```diff
@@ -1,154 +1,134 @@
...
 import * as React from 'react';
-import {View, StyleSheet} from 'react-native';
+import {ScrollView, View, StyleSheet, ImageBackground} from 'react-native';
 import {SafeAreaView} from 'react-navigation';
-import {NovaSolidLocationsLocationPinQuestionMark1 as QuestionIcon} from '@coorpacademy/nova-icons';
 
-import {SPACE} from '../const';
-import logo from '../assets/images/logo-coorp.png';
-import theme, {defaultHitSlop, BLUE_COORP_DARK, BLUE_COORP_LIGHT} from '../modules/theme';
+import background from '../assets/images/authentication.png';
```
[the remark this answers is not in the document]
VincentCATILLON: because it's a background

src/components/extralife.js
```diff
@@ -35,7 +35,7 @@ const styles = StyleSheet.create({
   },
   count: {
     fontWeight: theme.fontWeight.bold,
-    fontSize: theme.fontSize.extraLarge,
+    fontSize: theme.fontSize.xlarge,
```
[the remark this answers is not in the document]
VincentCATILLON: this is the butterfly effect lol

src/modules/theme.js
VincentCATILLON: because I previously changed the `Space` component with a switch case, and it was hurting me, and after, I changed for something lighter so you don't see it ^^

src/components/authentication.js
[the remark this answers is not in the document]
VincentCATILLON: is this blocking for you ?

VincentCATILLON:
@Volubyl Can you tell me what is blocking for you ? I'll make the updates but the PR is ready at this time :)

src/components/authentication-details.js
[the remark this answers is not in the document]
VincentCATILLON: Hm :/ For me, it's a bad idea to use directly the value from another file that is not related to the component you're using into your screen (the status bar color depends on their children background), and so, it's better to reuse color from their children instead of supposing they're using the potential color. It's true at T time, but not later if the child integration changes

PR #282, opened 2019-05-06
handle jwt video

Volubyl:
The JWT player video case was not handled before. This PR fix that mess.

### Result and observation

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


src/modules/media.js
```diff
@@ -1,14 +1,21 @@
...
 
-export const getMediaType = (mediaItem: Media): LessonType | void => {
+export const getMediaType = (mediaItem: Media | Lesson): LessonType | void => {
   return mediaItem.type;
 };
-export const getMediaUrl = (mediaItem: Media): string | void => {
+export const getMediaUrl = (mediaItem: Media | Lesson): string | void => {
   if (getMediaType(mediaItem) === RESOURCE_TYPE.VIDEO) {
+    if (mediaItem.videoId) {
+      // here flow cannot determine for sure if the mediaItem is either  a mediaItem or a Lesson.
```
Fausto95: nice

PR #285, opened 2019-05-09
Fix swipe hint animation: added resetAnimation

chrisdugne:
### before

<p align="center">
  
</p>

### after

<p align="center">
  
</p>

### Testing Strategy

- [x] Manual testing


chrisdugne:
[mobile animation : ETQU je veux savoir que je peux swiper les cards dans la popin interquestion](https://trello.com/c/kzpDezef/1167-mobile-animation-etqu-je-veux-savoir-que-je-peux-swiper-les-cards-dans-la-popin-interquestion)

src/containers/cards-swipable.js
```diff
@@ -98,24 +98,25 @@ class CardsSwipable extends React.PureComponent<Props, State> {
 
     this.setState({cardIndexShown: cardIndexSwiped + 1});
     this.hint.stop();
+    this.animation.resetAnimation();
```
chrisdugne: fix is here

VincentCATILLON: why don't have a start() and a stop() ? instead of having an animation loop that uses CPU for anything ?

PR #286, opened 2019-05-09
using tag for version

chrisdugne:
`Version: v1.2.3.commit`

Screen Shot 2019-05-09 at 5 13 56 PM


chrisdugne:
[ETQU j'aimerais le numéro de version de l'app et pas le numéro du commit](https://trello.com/c/yAZgXXHu/1275-etqu-jaimerais-le-num%C3%A9ro-de-version-de-lapp-et-pas-le-num%C3%A9ro-du-commit)

VincentCATILLON:
Can you please add a gif ?

VincentCATILLON:
As seen with you, we should concatenate the tag with the commit to be able to compare versions in internal store

src/components/version.js
```diff
@@ -15,7 +15,8 @@ type Props = {|
 const styles = StyleSheet.create({
   version: {
     fontSize: theme.fontSize.extraSmall,
-    textAlign: 'center'
+    textAlign: 'center',
+    paddingBottom: theme.spacing.medium
```
VincentCATILLON: should be added by home. sorry

VincentCATILLON:
GJ :)

PR #292, opened 2019-05-16
Fix dead end not resolved in Android

VincentCATILLON:
## Detailed purpose of the PR

This PR resolves the dead end in Android when no content has been found.
It also fixes a glitch that change the modal UI before it's hidden.

### Result and observation

Kapture 2019-05-16 at 14 16 29

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing


src/redux/reducers/ui/error.js
```diff
@@ -29,9 +29,7 @@ const reducer = <T>(state: State<T> = initialState, action: Action<T>): State<T>
     case HIDE: {
       return {
         ...state,
-        isVisible: false,
-        errorType: undefined,
-        lastAction: undefined
```
VincentCATILLON: This thing was making a modal UI change during hide, so bad.

src/layer/data/cards.js
```diff
@@ -163,11 +163,15 @@ const saveDashboardCardsInAsyncStorage = async (
...
-  const _cards = cardsToKeys(cards, language);
-  try {
-    await AsyncStorage.multiSet(cardsToPairs(_cards));
-  } catch (e) {
-    throw new Error('could not store the dashboard cards');
+  if (cards.length > 0) {
+    const _cards = cardsToPairs(cardsToKeys(cards, language));
+    if (_cards.length > 0) {
+      try {
+        await AsyncStorage.multiSet();
```
Fausto95: Shouldn't you save the `_cards` into the AS?

VincentCATILLON: only if there are cards to persist (the if statement 2 lines above)

Fausto95:
And why is the modal showing twice once you close it?

src/layer/data/cards.js
VincentCATILLON: Same thing than: https://github.com/facebook/react-native/issues/19163

VincentCATILLON:
@Fausto95 this is a lag due to my emulator RAM, it's not showed twice. it's only laggy

src/redux/actions/ui/modal.js
```diff
@@ -24,25 +24,28 @@ export const showModal = <T>({
...
+  });
 
-export const refresh = <T>(): StoreAction<Action<T>> => {
-  return (dispatch, getState) => {
-    const {error} = getState();
-    if (error.lastAction) {
-      dispatch(error.lastAction());
-    }
+export const refresh = <T>(): StoreAction<Action<T>> => async (dispatch, getState) => {
+  const {error} = getState();
```
Volubyl: to me you should'nt  do this because this action is only resposible to dispatch the last action without checking if the action exist. When using the refresh action you should assume that the last action exist.

So the check, should be done earlier.

src/redux/actions/cards.js
```diff
@@ -126,20 +126,20 @@ export const fetchCards = (
...
-            errorType: ERROR_TYPE.NO_CONTENT_FOUND,
-            lastAction: () => fetchCards(language)
-          })
-        );
+        // $FlowFixMe callable signature
+        return showModal({
+          errorType: ERROR_TYPE.NO_CONTENT_FOUND,
+          lastAction: () => fetchCards(language)
+          // $FlowFixMe union type
+        })(dispatch);
```
Volubyl: why adding dispatch ?

VincentCATILLON: to chain actions and avoid some crappy things: https://github.com/CoorpAcademy/mobile/pull/292/files#diff-8052d8a0c8d71b3103f140e0e4ddec62R37

src/redux/actions/ui/modal.js
VincentCATILLON: refresh is not responsible to hide the modal, it's not the action it was called for.

Btw, I'd to remove the hide action to be able to chain: https://github.com/CoorpAcademy/mobile/pull/292/files#diff-8052d8a0c8d71b3103f140e0e4ddec62R37

src/layer/data/cards.js
```diff
@@ -163,11 +163,15 @@ const saveDashboardCardsInAsyncStorage = async (
   cards: Array<DisciplineCard | ChapterCard>,
   language: SupportedLanguage
 ): Promise<void> => {
-  const _cards = cardsToKeys(cards, language);
-  try {
-    await AsyncStorage.multiSet(cardsToPairs(_cards));
-  } catch (e) {
-    throw new Error('could not store the dashboard cards');
+  if (cards.length > 0) {
```
```javascript 
  if (cards.length > 0) 
      try {
        const _cards = cardsToPairs(cardsToKeys(cards, language));
        await AsyncStorage.multiSet(_cards);
      } catch (e) {
        throw new Error('could not store the dashboard cards');
      }
  }
```

VincentCATILLON: done

VincentCATILLON:
Closed in favor of #293

PR #293, opened 2019-05-16
Fix error in android and modal UI change before its hidden

VincentCATILLON:
## Detailed purpose of the PR

This PR resolves the dead end in Android when no content has been found.
It also fixes a glitch that change the modal UI before it's hidden.

### Result and observation

No content case:

Kapture 2019-05-16 at 14 42 40

Just to be sure, forbidden access:

Kapture 2019-05-16 at 14 40 58

- [ ] **Breaking changes ?**
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [x] Unit testing


src/layer/data/cards.js
```diff
@@ -163,11 +163,15 @@ const saveDashboardCardsInAsyncStorage = async (
   cards: Array<DisciplineCard | ChapterCard>,
   language: SupportedLanguage
 ): Promise<void> => {
-  const _cards = cardsToKeys(cards, language);
-  try {
-    await AsyncStorage.multiSet(cardsToPairs(_cards));
-  } catch (e) {
-    throw new Error('could not store the dashboard cards');
+  if (cards.length > 0) {
+    const _cards = cardsToPairs(cardsToKeys(cards, language));
```
```javascript
  if (cards.length > 0) {
	      try {
	        const _cards = cardsToPairs(cardsToKeys(cards, language));
	        await AsyncStorage.multiSet(_cards);
	      } catch (e) {
	        throw new Error('could not store the dashboard cards');
      }
}
```

PR #295, opened 2019-05-16
segregate signin with QR Code from magic link

Volubyl:
## Detailed purpose of the PR

PRODUCT : The need is to be able to distinguish a signing using the QR Code from a signin through the magic link.

TECH : just pass a Authorization type to the signing redux action and GTM will do the rest.

HOW TO TEST :

*QRCODE* :
- Grab a test phone
-  log on your firebase account
- go to the debug view
- open your favourite web platform and scan the qr code
- then you should see the signin event on firebase with the right `authenticationType`

*MAGIC-LINK* : 

- same as QRCODE but you will need to use the magic link authentication method :-)

### Result and observation

LOOK at the `AuthenticationType` => qr code :-)

Screenshot 2019-05-16 at 18 27 14

And here look we've signed in using the magic link 

Screenshot 2019-05-16 at 18 31 22

- [ ] **Breaking changes ?**  

- [ ] **Extra dependency ?**

### Testing Strategy

- [X] Already covered by tests
- [ ] Manual testing
- [X] Unit testing


src/screens/authentication.js
```diff
@@ -75,13 +76,13 @@ class AuthenticationScreen extends React.PureComponent<Props, State> {
 
   handleSignOut = () => this.props.navigation.popToTop();
 
-  handleSignIn = async (token?: string) => {
-    await this.props.signIn(token);
+  handleSignIn = async (authenticationType: ?AuthenticationType, token?: string) => {
```
VincentCATILLON: should throw a flow error.

src/redux/actions/authentication.js
```diff
@@ -64,11 +65,10 @@ export const getAnonymousToken = async (): Promise<string> => {
   return token;
 };
 
-export const signIn = (_token?: string): StoreAction<Action | BrandsAction> => async (
-  dispatch,
-  getState,
-  options
-) => {
+export const signIn = (
+  authenticationType: AuthenticationType,
```
VincentCATILLON: `authenticationType: AuthenticationType | null`

PR #301, opened 2019-05-21
Force Roboto font for One Plus users

VincentCATILLON:
## Detailed purpose of the PR

Related issue: https://github.com/facebook/react-native/issues/15114

### Result and observation

You should use the app with non truncated texts.

- [ ] **Breaking changes ?**  
- [ ] **Extra dependency ?**

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing


VincentCATILLON:
Need functional review to see if the problem still persist.

VincentCATILLON:
Tested with a One Plus, and it works now :)

PR #314, opened 2019-06-03
Plugging sections

chrisdugne:
<p align="center">
  
</p>

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


VincentCATILLON:
IMO we should distinguish content by language

src/containers/catalog.js
```diff
@@ -90,10 +93,9 @@ class Catalog extends React.PureComponent<Props, State> {
 
 const mapStateToProps = ({catalog, authentication, ...state}: StoreState): ConnectedStateProps => {
   const language = translations.getLanguage();
-  const {brand} = authentication;
 
   return {
-    dashboardSections: brand && brand.dashboardSections,
+    dashboardSections: catalog.entities.sections,
```
VincentCATILLON: should have the same structure than cards and be reduced (for languages support)

src/redux/reducers/catalog.js
```diff
@@ -17,7 +17,7 @@ export type State = {|
     },
     sections: {
       [key: string]: {
-        ...DashboardSection,
+        ...Section,
```
VincentCATILLON: This extended type should be defined (and exported) to be reused in some containers (catalog for example)

VincentCATILLON: So for this, this type could be defined and named `Section`, and the other type (currently `Section`) could be named `ApiSection` to be explicit

src/containers/catalog.js
```diff
@@ -8,18 +8,20 @@ import type {Props as ComponentProps} from '../components/catalog';
...
 import {fetchCards} from '../redux/actions/catalog';
+import {fetchSections} from '../redux/actions/sections';
 import translations from '../translations';
 import {compareCards} from '../utils/content';
-import type {DashboardSection} from '../types';
+import type {Section} from '../types';
 
 type ConnectedStateProps = {|
-  dashboardSections: Array<DashboardSection>,
+  dashboardSections: Array<Section>,
```
VincentCATILLON: And this one should be the new `Section` type (not the `ApiSection`)

src/containers/catalog.js
```diff
@@ -44,13 +45,11 @@ class Catalog extends React.PureComponent<Props, State> {
...
-    return this.props.dashboardSections;
-  };
-
   fetchContent = async () => {
-    const visibleSections = this.getVisibleSections();
-    await this.props.fetchCards(visibleSections, translations.getLanguage());
+    const {sections} = this.props;
+    await this.props.fetchSections(translations.getLanguage());
+    // @todo map on sections
+    await this.props.fetchCards(sections, translations.getLanguage());
```
Fausto95: This is not really clear. If I'm going to read the code, I'll expect sections to be returned from fetchSections.
So what does this fetchSections does?

VincentCATILLON: don't take care of this, it will be moved elsewhere, it's a legacy code that is not updated yet

PR #315, opened 2019-06-03
Persist bundle

VincentCATILLON:
## Detailed purpose of the PR

This PR persists bundle state (remember, only the keys are stored in this one) and rehydrates it at start to be synchronized with offline elements previously stored in async storage.

### Result and observation

When no bundle has been fetched yet:

Capture d’écran 2019-06-03 à 16 37 34

When bundled was previously fetched:

Capture d’écran 2019-06-03 à 16 40 53

- [ ] **Breaking changes ?**
- [x] **Extra dependency ?** `redux-persist@5` that we'll be able to use with `redux-offline` once we'll start discussing on it 

### Testing Strategy

- [ ] Manual testing
- [ ] Unit testing

### DIFF
--- a/flow-typed/_custom-defs.js
+++ b/flow-typed/_custom-defs.js
@@ -38,3 +38,5 @@ declare type ReactNavigation$WithNavigationProps = $Exact<
 
 declare type Dispatch = Redux$Dispatch<*>;
 declare type GetState = () => StoreState;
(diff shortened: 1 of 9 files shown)


VincentCATILLON:
This is not needed anymore. We'll check in async storage directly.

PR #318, opened 2019-06-05
testing waitFor

chrisdugne:
this PR should NOT pass e2e tests.
e2e `waitFor` are to be investigated...


chrisdugne:
```
    NOTE: Every waitFor call must set a timeout using withTimeout(). Calling waitFor without setting a timeout will do nothing.
```

```
    NOTE: waitFor will not throw when reaching timeout, instead it will just continue to the next line. To make sure your tests work as you expect them to add expect() at the following line.
```
from https://github.com/wix/detox/blob/master/docs/APIRef.waitFor.md

PR #329, opened 2019-06-13
Fix correction card android overflow bug

Fausto95:
# Before 

Screenshot_20190617-120957

# After

Screenshot_20190613-100156


VincentCATILLON:
Fixed by https://github.com/CoorpAcademy/mobile/pull/324 for the bug you're introducing. you should hide overflow on card content and not the entire card to avoid this :)

src/components/lives.js
```diff
@@ -121,20 +121,27 @@ const Lives = ({
   };
 
   const brokenSuffix = isBroken ? '-broken' : '';
-  const topText = (isGodMode && '∞') || winningLife ? count - 1 : count;
+  const godModeInfiniteLoopChar = '∞';
```
VincentCATILLON: should be a constant IMO

Fausto95:
[Fix Card Correction Bug on Android(List of response(RN59))](https://trello.com/c/6Gr4kGOG/1311-fix-card-correction-bug-on-androidlist-of-responsern59)

PR #335, opened 2019-06-14
Fix touchable catalog item

Fausto95:
# Before
ezgif com-optimize (1)

# After 
ezgif com-optimize (2)


Fausto95:
Yeah, it does!

PR #336, opened 2019-06-17
Use no-JIT JSC

Fausto95:
Fix to use JSC-only packageron android and remove useDeprecatedNdk filters


Volubyl:
for fred, with love <3

PR #337, opened 2019-06-17
Dont use default react-native's gradle

Fausto95:
https://github.com/facebook/react-native/issues/24505


Fausto95:
This is to avoid app crashing on release mode!

PR #341, opened 2019-06-18
Remove Refresh Control due to its integration on flatlist

Fausto95:
Removing RefreshControl component due to its integration into FlatList


VincentCATILLON:
@Volubyl @chrisdugne

PR #342, opened 2019-06-18
Fix Storybook

Fausto95:
Bring Storybook back to life
Fix Storybook due to metro config change between RN 0.56 -> 0.59
`blacklistRE` is no longer a callable `function` :)

And I know you might be wondering why not do this 

```javascript 
const blacklist = () => {
  if (REACT_NATIVE_FLAVOR === 'STORYBOOK') {
    // this is to have fixtures embedded in storybook app
    return createBlacklist([]);
  }
};

blacklistRE: blacklist()
``` 

instead of

```javascript
    blacklistRE: (() => {
      if (REACT_NATIVE_FLAVOR === 'STORYBOOK') {
        // this is to have fixtures embedded in storybook app
        return createBlacklist([]);
      }
    })()
```

I just love IIFEs that's it :)


VincentCATILLON:
Noooo, kidding ??? 😁

Fausto95:
it goes without saying :)

PR #343, opened 2019-06-18
Feedback sonore

Fausto95:
Feedback sonore [WIP]


src/containers/audio-player.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+    sound.play(() => {
+      sound.release();
+    });
+  };
+
+  const _player = new Sound(soundFile, error => callback(error, _player));
+};
+
+export const playWrongAnswerSound = () => Player(require('../assets/sounds/wrong-answer.mp3'));
```
Volubyl: I wonder if those files should'nt be stored in a S3 bucket because storing such assets may make the app heavier

VincentCATILLON: But it won't be reachable in offline mode, and the files are really smalls

src/containers/audio-player.js
```diff
@@ -0,0 +1,29 @@
...
+export const playGoodAnswerSound = () => Player(require('../assets/sounds/good-answer.mp3'));
+export const playSuccessLevelEndSound = () =>
+  Player(require('../assets/sounds/success-level-end.mp3'));
+export const playFailureLevelEndSound = () =>
+  Player(require('../assets/sounds/failure-level-end.mp3'));
+
+export default Player;
```
VincentCATILLON: should be in modules, it's not a container as it doesn't contain anything

src/screens/correction.js
```diff
@@ -119,6 +120,12 @@ class CorrectionScreen extends React.PureComponent<Props, State> {
     }
   });
 
+  componentDidMount() {
+    const {isCorrect} = this.props;
+    if (isCorrect) return playGoodAnswerSound();
```
VincentCATILLON: Add brackets please

src/screens/level-end.js
```diff
@@ -59,6 +60,12 @@ class LevelEndScreen extends React.PureComponent<Props, State> {
     headerStyle: {}
   });
 
+  componentDidMount() {
+    const {isCorrect} = this.props.navigation.state.params;
+    if (isCorrect) return playSuccessLevelEndSound();
```
VincentCATILLON: Add brackets please

src/containers/audio-player.js
Fausto95: I know I know, just put it in here because I dunno how to test it yet :)

src/screens/correction.js
Fausto95: Why?

VincentCATILLON: Yes it's not readable

src/containers/audio-player.js
VincentCATILLON: so please move it in helpers and we'll add some tests in R&D tasks, don't worry dude

src/modules/audio-player.js
```diff
@@ -0,0 +1,37 @@
...
+
+const playFailureLevelEndSound = () =>
+  AudioPlayer(require('../assets/sounds/failure-level-end.mp3'));
+
+export default {
+  playWrongAnswerSound,
+  playGoodAnswerSound,
+  playSuccessLevelEndSound,
+  playFailureLevelEndSound
+};
```
VincentCATILLON: can we not imagine a generic function (ie: `playSound`) that takes in parameter the file name (not the entire path and instead of requiring it) that makes it for every file provided ?

```
const SOUNDS_PATH = path.resolve([...]);

export const playSound = (name: string) => {
  Sound.setCategory('Playback');

   const callback = (error, sound) => {
    if (error) {
      return;
    }
    sound.play(() => {
      sound.release();
    });
  };

   const _player = new Sound(require(`${SOUNDS_PATH}/name`), error => callback(error, _player));
};
```

or, if it's not possible, to give directly the required file in params (as you're doing it there), but exposed only this function instead of exposing 1 function for each file you'll use ?

src/modules/audio-player.js
```diff
@@ -0,0 +1,37 @@
+/* istanbul ignore file */
+
+// @flow
+
+import Sound from 'react-native-sound';
+
+const AudioPlayer = (soundFile: File) => {
+  Sound.setCategory('Playback');
```
chrisdugne: no need to `Sound.setCategory` each time you `AudioPlayer()` ?

src/modules/audio-player.js
chrisdugne: if so, you need to provide a `Sounds` type to map the available names.

src/modules/audio-player.js
```diff
@@ -4,9 +4,15 @@
 
 import Sound from 'react-native-sound';
 
-const AudioPlayer = (soundFile: File) => {
-  Sound.setCategory('Playback');
-
+Sound.setCategory('Playback');
+
+const AudioPlayer = (soundName: string) => {
```
VincentCATILLON: add a type on this please

VincentCATILLON: can you please rename it for something like `playSound` ? Btw it's not a class, it should be something link `audioPlayer`

src/modules/audio-player.js
```diff
@@ -4,9 +4,15 @@
...
+  const sounds = {
```
VincentCATILLON: add a type on this constant, and define it elsewhere than in the function please

PR #353, opened 2019-06-25
updated PR template to match mooc's one

chrisdugne:
## Detailed purpose of the PR

> updated PR template to become this one

### Testing strategy


VincentCATILLON:
The previous template was better IMO but ok

PR #354, opened 2019-06-25
Handle empty sections

VincentCATILLON:
## Detailed purpose of the PR

This PR is the next step of removing sections filter on back, we have to do it in front.

### Result and observation

The sections fetch should be faster than before and you should never see empty sections.

- [ ] **Breaking changes ?** 
- [ ] **Extra dependency ?**

### Testing Strategy

- [x] Manual testing
- [x] Unit testing


src/containers/catalog.js
```diff
@@ -52,16 +54,33 @@ class Catalog extends React.PureComponent<Props, State> {
...
+  }
+
+  fetchSections = async (offset: number, limit: number, forceRefresh?: boolean = false) => {
+    await this.props.fetchSections(offset, limit, translations.getLanguage(), forceRefresh);
   };
 
   handleRefresh = () => {
     this.setState({isRefreshing: true});
-    this.fetchSections(0, DEFAULT_LIMIT)
+    this.fetchSections(0, this.getLimit(0), true)
```
VincentCATILLON: `true` is to force the refresh

src/containers/catalog.js
```diff
@@ -52,16 +54,33 @@ class Catalog extends React.PureComponent<Props, State> {
   offsetY: number = 0;
 
   componentDidMount() {
-    this.fetchSections(0, DEFAULT_LIMIT);
+    this.fetchSections(0, this.getLimit(0));
   }
 
-  fetchSections = async (offset: number, limit: number) => {
-    await this.props.fetchSections(offset, limit, translations.getLanguage());
+  componentDidUpdate(prevProps: Props) {
```
VincentCATILLON: this part is to handle empty sections reduced in state and retrigger a fetch if any unfetched section is visible.

PR #363, opened 2019-07-03
Fixes the calculated offset depending on hidden sections

VincentCATILLON:
## Detailed purpose of the PR

This PR fixes the following issues:
- the calculated offset was wrong in case of multiple chained empty sections

### Result

**Before**:

Kapture 2019-07-03 at 14 32 16

**After**:

Kapture 2019-07-03 at 14 29 48

### Testing strategy


src/modules/sections.test.js
```diff
@@ -25,6 +25,39 @@ describe('Sections', () => {
...
+
+    it('should increase the offset to bypass empty sections', () => {
+      const {getOffsetWithoutCards} = require('./sections');
+      const sections = [emptySection, emptySection, undefined, undefined, emptySection, undefined];
+      const result = getOffsetWithoutCards(sections, 0);
+      const expected = 2;
+      expect(result).toEqual(expected);
+    });
+
+    it('should increase the offset to bypass empty sections', () => {
```
VincentCATILLON: i should change this test label (duplicate)

PR #366, opened 2019-07-04
Remove duplicate color

VincentCATILLON:
## Detailed purpose of the PR

This PR removes duplicate color from theme.

### Testing strategy


src/modules/theme.js
```diff
@@ -72,8 +71,7 @@ const theme: Theme = {
     black: '#14171A',
     battle: '#FFE100',
     notification: '#FF7043',
-    salmon: '#fde2e5',
-    pink: '#f73f52'
```
VincentCATILLON: duplicate of negative one

PR #369, opened 2019-07-08
Fix/language

Volubyl:
## Detailed purpose of the PR

Two bugs :

1. The language was wrongly provided to the sections API and that one was only returning sections english based

2. The library used to get the device language (react-native-localization) has been introduced an issue in his most recent version (`2.1.2`) : the method `getLanguage` was not returning the real locale.

=> so we downgrade the version to an older but stable one 

### Testing strategy

- [X] I ensured my code is tested:  UT and by ✋ 


src/layer/data/sections.js
```diff
@@ -25,14 +27,18 @@ export const fetchSections = async (
...
-      headers: {
-        authorization: token
-      }
-    }
-  );
+  const query = {
+    type: 'cards',
+    offset,
+    limit,
+    withoutAdaptive: true,
```
VincentCATILLON: should be removed

src/layer/data/sections.test.js
```diff
@@ -52,7 +52,9 @@ describe('sections', () => {
             hits: Array<Section>
           |}>
         }> => {
-          expect(url).toBe('host/api/v2/sections?offset=1&limit=2&type=cards');
+          expect(url).toBe(
+            'host/api/v2/sections?type=cards&offset=1&limit=2&withoutAdaptive=true&lang=fr'
```
VincentCATILLON: withoutAdaptive should be removed

src/modules/uri.js
```diff
@@ -20,3 +20,16 @@ export const getResourceUrl = (resource: Lesson): string | void => {
 
   return url;
 };
+
+type QueryParams = {
```
VincentCATILLON: strict type please

src/modules/uri.js
```diff
@@ -20,3 +20,16 @@ export const getResourceUrl = (resource: Lesson): string | void => {
...
+};
+
+export const buildUrlQueryParams = (params: QueryParams) =>
+  Object.keys(params)
+    .map(key => {
+      const value = (params[key] !== undefined && params[key]).toString() || '';
+
+      return `${encodeURIComponent(key)}=${encodeURIComponent(value)}`;
+    })
+    .join('&');
```
VincentCATILLON: love it thx

src/layer/data/sections.js
```diff
@@ -25,14 +27,18 @@ export const fetchSections = async (
...
 
-  const response = await fetch(
-    `${jwt.host}/api/v2/sections?offset=${offset}&limit=${limit}&type=cards`,
-    {
-      headers: {
-        authorization: token
-      }
-    }
-  );
+  const query = {
```
VincentCATILLON: use QueryParams type you defined in modules/uri

src/modules/uri.js
```diff
@@ -20,3 +20,16 @@ export const getResourceUrl = (resource: Lesson): string | void => {
 
   return url;
 };
+
+type QueryParams = {
+  [key: string]: string | number
```
VincentCATILLON: add boolean please :)

PR #372, opened 2019-07-17
coverage-select-card

chrisdugne:
## Detailed purpose of the PR

coverage select card

### Testing strategy

- [x] I ensured my code is tested: 
- [x] I took care to mention breaking changes and extra libs if appropriate

PR #373, opened 2019-07-17
Fix next level on level end screen and completion update

VincentCATILLON:
## Detailed purpose of the PR

Fix next level action on level end screen.

To understand, previously the next level was calculated from card completion in the level-end (which is player scope related), so, as we were unable to find the reason of the bug, we changed this logic to use dedicated service functions (`getNext`, `getNextChapter` and `getNextLevel`), to be ISO with the web implementation (mooc, not player-web).

Btw, this PR fixes another spotted issue about completions not updated once back to home (since dashboard has been done).

NB: There is a refacto on cards redux actions to have something more little to test (there was a big file, and as we changed the `selectCard` action, we had to update all tests around it)

### Result

Kapture 2019-07-18 at 16 23 53

### Testing strategy

- [x] I ensured my code is tested: unit, storyshots


VincentCATILLON:
Waiting for https://github.com/CoorpAcademy/mobile/pull/372

jest.config.unit.js
```diff
@@ -36,14 +36,14 @@ module.exports = {
     'src/redux': {
       branches: 88,
       function: 100,
-      lines: 100,
-      statements: 100
+      lines: 95,
+      statements: 95
```
VincentCATILLON: Ticket created to go back to 100% after meeting about expedits action plan.

chrisdugne: https://github.com/CoorpAcademy/mobile/tree/coverage-create-next-progression

PR #378, opened 2019-07-22
Fix Circle CI Build

Fausto95:
This PR: 
`bumps every circleci version cache`;
`bumps xcode version`;


Volubyl:
Seems supa gr8! Waiting another review from something more comfy about CI conf like @AdrieanKhisbe  ?

.circleci/config.yml
```diff
@@ -147,20 +147,20 @@ jobs:
               brew link applesimutils
             else
               echo "Applesimutils not installed."
-              brew tap wix/brew
+              brew untap wix/brew && brew tap wix/brew
```
Volubyl: Could you explain this line ? why untapping and tapping ?

Fausto95: This is just to make sure there are no conflits;

From detox docs:
```
If Homebrew complains about a conflict in the wix/brew tap, run brew untap wix/brew && brew tap wix/brew and try installing again
``` 
https://github.com/wix/AppleSimulatorUtils

.circleci/config.yml
```diff
@@ -147,19 +147,21 @@ jobs:
...
+          key: brew-e2e-v5-ios-{{ arch }}
           paths:
             - ~/Library/Caches/Homebrew
             - /usr/local/Cellar
 
       - run:
           name: Tests
           command: |
+            yarn global add detox-cli
+            detox clean-framework-cache && detox build-framework-cache
```
Fausto95: Had to do this here because detox seemed to not be installed in this step
This is one of the fixes

.circleci/config.yml
AdrieanKhisbe: ✨

PR #381, opened 2019-07-25
Use everywere the app language as single source of truth

Volubyl:
## Detailed purpose of the PR

The main idea : using the app language as the Single and Truthful  source of truth!

**Why** ?

Before : we were using the content locale (`card.lang`) to store data ( and to build the storage key like `foo:en:bar` => if the content was in english) but we were using the mobile language to get the data after creating the key. 

And if the content was in `en` but the app in `cs` the key were respectively `foo:en:bar` and `foo:cs:bar` => so 💥 💥  KABOOOOOM 💥 💥  the mighty `no content found` modal showed up!

So now, we should use the same locale everytime for write and read operation and rely on the API to returns the content with the accurate language.

** How to reproduce ? **

- First use your mobile phone logged with `onboarding` (will work with others platforms but  it's  just more straightforward to explain using `onboarding` as example )

- Set your phone un Czech (Tchèque -- again it will work also with other unsupported language ...)

- Go back to the app and click on whatever card you like

- Now should the the question written in english but button in Czech

### Testing strategy

- [X] I ensured my code is tested: 
- [X] I took care to mention breaking changes and extra libs if appropriate


Volubyl:
code coverage increase in progress

src/layer/data/index.js
```diff
@@ -27,8 +30,13 @@ import {fetchSections} from './sections';
 
 export type DataLayer = {
   ...DataLayerBase,
-  fetchBundle: typeof fetchBundle,
-  storeBundle: typeof storeBundle,
+  fetchBundle: (
```
godu: Utilise [$Call](https://flow.org/en/docs/types/utilities/#toc-call)

PR #391, opened 2019-08-14
Use only one component for images and use media API to resize them

VincentCATILLON:
## Detailed purpose of the PR

This PR uses media API to optimize usage (by resize) of images everywhere.

Btw, we only use 1 image component right now, instead of 4 different before: 
- `ImageGradient` (local)
- `ImageBackground` (react-native)
- `Image` (local)
- `Image` (react-native)

### Testing strategy

- [x] I took care to mention breaking changes and extra libs if appropriate: `Image` and `ImageGradient` components have been removed


VincentCATILLON:
Idea: use a component instead of a helper

src/modules/image.js
```diff
@@ -31,3 +35,22 @@ export const getImageDimensions = (
...
+export const getResizedImage = (
+  url: string,
+  {maxWidth, maxHeight, resizeMode}: {maxWidth?: number, maxHeight?: number, resizeMode?: 'contain'}
+) => {
+  if (!maxWidth && !maxHeight) {
+    return url;
+  }
+
+  const queryParams = buildUrlQueryParams({
+    ...(maxWidth ? {w: maxWidth} : {}),
```
Fausto95: these spreads could be turned into something more readable

VincentCATILLON: I agree with you, i'll make a component and move this logic into this one and make it easier to read btw

src/components/resources-browser.js
```diff
@@ -97,37 +100,18 @@ class ResourcesBrowser extends React.PureComponent<Props> {
...
-                  )}
+                  <Preview
+                    type={resource.type}
+                    source={{uri: resource.poster}}
+                    hasOverlay={!isSelected}
+                    iconWidth={20}
+                    iconHeight={20}
+                    isIconVisible={!isSelected}
+                    style={styles.thumbnail}
+                  />
```
VincentCATILLON: This thing has been refactored due to duplicated code with `Preview` component

src/modules/image.js
```diff
@@ -1,33 +1,44 @@
 // @flow strict
 
-export type Dimensions = {|
-  width: number,
-  height: number
-|};
-
-export const getImageDimensions = (
```
VincentCATILLON: This function has been removed because it's unused now

src/modules/image.js
VincentCATILLON: done @Fausto95 :) https://github.com/CoorpAcademy/mobile/pull/391/files?file-filters%5B%5D=.js#diff-62408838b1f3a42b337cd185b30fd9b3R21

src/components/button-sticky.stories.js
```diff
@@ -6,7 +6,7 @@ import {storiesOf} from '@storybook/react-native';
 import {handleFakePress} from '../utils/tests';
 import {Component as ButtonSticky} from './button-sticky';
 
-storiesOf('Button Footer', module)
+storiesOf('ButtonSticky', module)
```
VincentCATILLON: Story name not up to date

src/components/catalog.stories.js
```diff
@@ -19,7 +19,11 @@ if (__TEST__) {
 const sections = createSections();
 const sectionsWithCardsRef = sections.map((section, index) => ({
   ...section,
-  cardsRef: (index === 0 && ['foo']) || (index === 1 && ['bar']) || (index === 2 && []) || undefined
+  cardsRef:
+    (index === 0 && ['foo', 'bar']) ||
+    (index === 1 && ['bar', 'foo']) ||
+    (index === 2 && []) ||
+    undefined
```
VincentCATILLON: just to see multiple cards next to each other

src/components/brand-theme-provider.stories.js
```diff
@@ -17,7 +17,11 @@ storiesOf('BrandThemeProvider', module).add('Default', () => (
             <Text style={{color: brandTheme.colors.primary}}>Primary color</Text>
             <Text>Content category name: {brandTheme.contentCategoryName}</Text>
             <Text>Name: {brandTheme.name}</Text>
-            <Image source={{uri: brandTheme.images['logo-mobile']}} style={{width: 200}} />
+            <Image
```
chrisdugne: ImageBackground ?

VincentCATILLON: I can use it yes, but it's only to see brand theme context, so it's not really important

VincentCATILLON: IMO, we should use a text component to display the value instead of the image

PR #396, opened 2019-08-22
Dashboard render optimization

VincentCATILLON:
## Detailed purpose of the PR

This PR improves the props comparison in catalog section component to `PureComponent` shallow compare method poor performances on huge props and adds memoize on `mapStateToProps` (on state and on lists items with `reselect-map`).

## Result

The test was made in similar conditions (initial render, scrolling to the longest section, scrolling 25 times on the right to trigger fetching and re-renders, and select a card).

**We reduced about ~82% of useless renders.**

Before: `873`
After: `158`

NB: I noticed 10MB memory usage reduce after this optimization (and the UI never drops under 59fps).

### Testing strategy

- [x] I took care to mention breaking changes and extra libs if appropriate: `reselect` & `reselect-map` (for lists)

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,6 +82,7 @@
     "hoist-non-react-statics": "^3.2.1",
     "jsc-android": "^241213.2.0",
     "jwt-decode": "^2.2.0",
+    "lodash": "^4.17.15",
     "p-map": "^2.1.0",
     "react": "16.8.3",
     "react-native": "0.59.9",
--- a/package.json
+++ b/package.json
@@ -113,6 +113,8 @@
     "react-redux": "^7.0.3",
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
+    "reselect": "^4.0.0",
+    "reselect-map": "^1.0.4",
     "rn-fetch-blob": "0.10.15",
     "rn-placeholder": "^2.0.0",
     "semver": "^6.1.1"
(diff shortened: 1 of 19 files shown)


VincentCATILLON:
@godu this is another subtask (using reselect), this is to fix the shallow compare on huge props which force re-renders on `PureComponent`

godu:
Alors cette tache ne sert à rien. Si le memoize est fait plus haut. ce dev rajoute juste un check inutile.

VincentCATILLON:
🤔  Hm, I see what you means. I'll check once mapStateToProps memoized to see if this dev has any added value or if it's something that we can drop

src/modules/equal.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+import isEqual from 'lodash/fp/isEqual';
```
VincentCATILLON: Used to have a really shallow compare

Volubyl:
no test for the map state to props ?

VincentCATILLON:
@Volubyl you're right, i'll add these ones

VincentCATILLON:
I was hoping a `kaboom` from you 🙄 @godu

package.json
```diff
@@ -112,6 +113,8 @@
     "react-redux": "^7.0.3",
     "redux": "^4.0.1",
     "redux-thunk": "^2.3.0",
+    "reselect": "^4.0.0",
```
chrisdugne: this one is not used then

VincentCATILLON: Needed because of these dependencies management: https://github.com/heyimalex/reselect-map/blob/master/package.json#L29

PR #401, opened 2019-08-29
Testing progression bar

Fausto95:
This PR adds e2e tests to check whether the progressbar progression works after answering to question or either when we reload the app


__e2e__/progression.test.js
```diff
@@ -26,4 +28,36 @@ describe('Progression bar', () => {
...
+    beforeAll(async () => {
+      await reloadApp();
+    });
+
+    it('should save progress bar to its state before reloading the app', async () => {
+      await weExpect(
+        element(
+          by
+            .type('RCTView')
+            .and(by.id('progression-bar-0.08333333333333333'))
```
VincentCATILLON: why do you need to select by type in addition to id ? its not enough ?

Fausto95: Too nested

VincentCATILLON:
should be rebased on master to ensure the tests are working :)

PR #407, opened 2019-09-05
Apply pixel ratio to resized images

VincentCATILLON:
## Detailed purpose of the PR

This PR uses pixel ratio for every image we are resizing in media API, to avoid having some images in pixel art.

### Result

Based on iPhone XR ([Pixel ratio at 2](https://facebook.github.io/react-native/docs/pixelratio#get))

Before:

Capture d’écran 2019-09-05 à 13 33 32

After:

Capture d’écran 2019-09-05 à 13 33 44

### Testing strategy

- [x] I ensured my code is tested: unit tests + snapshots


src/modules/image.js
```diff
@@ -27,14 +28,14 @@ export const getResizedImage = (
   if (maxWidth) {
     queryParams = {
       ...queryParams,
-      w: maxWidth
+      w: PixelRatio.getPixelSizeForLayoutSize(maxWidth)
```
VincentCATILLON: Source: https://facebook.github.io/react-native/docs/pixelratio

src/modules/image.js
```diff
@@ -27,14 +28,14 @@ export const getResizedImage = (
...
     };
   }
 
   if (maxHeight) {
     queryParams = {
       ...queryParams,
-      h: maxHeight
+      h: PixelRatio.getPixelSizeForLayoutSize(maxHeight)
```
VincentCATILLON: Source: https://facebook.github.io/react-native/docs/pixelratio

PR #408, opened 2019-09-05
Memoize and unit tests on screens mapStateToProps

VincentCATILLON:
## Detailed purpose of the PR

Add memoize and unit tests on all screens mapStateToProps

### Testing strategy

- [ ] I ensured my code is tested: unit tests


VincentCATILLON:
Missing things:
- Lesson resource type is not as expected (?)
- Memoize on Correction and Question to be done

jest.config.unit.js
```diff
@@ -35,21 +35,21 @@ module.exports = {
       // @todo increase this value
       branches: 40,
       // @todo increase this value
-      functions: 42,
+      functions: 41,
       // @todo increase this value
       lines: 46,
       // @todo increase this value
-      statements: 47
+      statements: 46
```
VincentCATILLON: This value decreased because I moved some state extractor from containers to state extractor file, so the percentage decreased from 46.0x to 45.9x 🙄

jest.config.unit.js
```diff
@@ -35,21 +35,21 @@ module.exports = {
       // @todo increase this value
       branches: 40,
       // @todo increase this value
-      functions: 42,
+      functions: 41,
```
VincentCATILLON: This value decreased because I moved some state extractor from containers to state extractor file, so the percentage decreased from 42.0x to 41.9x 🙄

jest.config.unit.js
```diff
@@ -33,23 +33,23 @@ module.exports = {
     },
     'src/containers': {
       // @todo increase this value
-      branches: 40,
+      branches: 39,
```
VincentCATILLON: This value decreased because I lightened `question-slider` container, so the percentage decreased from 40.0x to 39.9x 🙄

jest.config.unit.js
Fausto95: Not sure about this

src/components/question-choices.js
```diff
@@ -138,13 +140,15 @@ class QuestionChoices extends React.PureComponent<Props> {
           </View>
         );
       case QUESTION_TYPE.SLIDER: {
-        if (!min || !max) {
+        if (min === undefined || max === undefined) {
```
Fausto95: Why this change?

Volubyl:
I'm really understand the @Fausto95  point : the coverage is very pretty low ...

src/components/question-choices.js
VincentCATILLON: Because the unit data was redundant, `Edge` type was useless so, and default value from `mapStateToProps` was an object containing properties with `"undefined undefined"` instead of undefined props, as simple as it should.

PR #419, opened 2019-09-17
Hero redux

VincentCATILLON:
## Detailed purpose of the PR

This PR adds everything around redux to fetch / store hero.

It also bumps the redux coverage to 100%

### Testing strategy

- [x] I ensured my code is tested: unit tests


VincentCATILLON:
Missing thing:
- state extract
- selector (to memoize)

src/redux/actions/progressions/create-chapter-progression.test.js
```diff
@@ -3,37 +3,70 @@
...
+  it('should create chapter progression with specific engine version', async () => {
+    const playerStore = require('@coorpacademy/player-store');
+    const {createChapterProgression} = require('./create-chapter-progression');
+
     const chapter = createChapter({
       ref: 'cha_1',
       name: 'chapter'
     });
 
-    const ENGINE_VERSION = '2';
```
Fausto95: IMO `Engine_Version` is more explicit

src/redux/actions/progressions/create-chapter-progression.test.js
```diff
@@ -3,37 +3,70 @@
...
     // $FlowFixMe
-    const actual = await createChapterProgression(chapter, ENGINE_VERSION);
+    const actual = await createChapterProgression(chapter, version);
 
     expect(playerStore.createProgression).toHaveBeenCalledTimes(1);
     expect(actual).toEqual({type: '@@mock/CREATE_PROGRESSION', payload: {_id: '__ID__'}});
   });
+
+  it('should create chapter progression with default engine version', async () => {
+    const playerStore = require('@coorpacademy/player-store');
```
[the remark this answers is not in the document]
VincentCATILLON: because thats the way we deal with unit tests, importing them in each test ensure that you have a proper module everytime, you can mock before, have a cleaner implementation (only once) and be sure that you don't have a previous mock on other test

src/redux/actions/progressions/create-chapter-progression.test.js
VincentCATILLON: but uppercase vars are for constants, there it's only a variable (even if the keyword is `const`)

src/redux/actions/progressions/create-chapter-progression.test.js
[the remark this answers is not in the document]
VincentCATILLON: btw, its not imported twice, its mocked [there](https://github.com/CoorpAcademy/mobile/pull/419/files/be1cc809f635807f1f7f9c3a8f26877034642446#diff-e079ef792880127c2b7a99b9dc6064faR7) and imported in each test to mock them independently

src/containers/catalog.js
```diff
@@ -178,13 +181,19 @@ class Catalog extends React.Component<Props, State> {
   }
 }
 
-const getSectionsState = createArraySelector(
+const getHeroState: StoreState => DisciplineCard | ChapterCard | void = createSelector(
```
chrisdugne: how about creating a `type Hero = DisciplineCard | ChapterCard` ?
then when features change and hero gets more props we need to change here only

VincentCATILLON: why not, shouldn't we do this once we need to do it ?

PR #420, opened 2019-09-17
Fix conflicts version and eslint config

VincentCATILLON:
## Detailed purpose of the PR

This PR tries to fix unit tests broken due to hooks usage.

### Testing strategy


Fausto95:
Can't understand how tests passed yesterday still

src/containers/with-analytics.js
```diff
@@ -3,13 +3,20 @@
 import * as React from 'react';
 import hoistNonReactStatic from 'hoist-non-react-statics';
 
-import {AnalyticsContext} from '../components/analytics-provider';
-import type {State as AnalyticsState} from '../components/analytics-provider';
+import type {Services} from '../services'
+import createServices from '../services';
+import createDataLayer from '../layer/data';
```
VincentCATILLON: this part is copy pasted from old analytics provider that i removed because it doesnt handle any state, so, its useless

PR #437, opened 2019-10-31
Fixing Correction Screen Error Answers

Fausto95:
This PR fixes a weird bug in the correction screen

After analyse, we figured out that the card component was not re-render when it should be because of the way props were passed.

So the fix is to pass the same props as before `userAnswers` & `answers` but at the right moment, that means that when they're hot so that way the component detects a new props change and then re-render the component

# Before 

Capture d’écran 2019-10-29 à 15 14 06

# After 

Capture d’écran 2019-10-29 à 15 14 06

Trello: https://trello.com/c/tqyE4RUh/1479-as-a-mobile-user-on-the-inter-question-popin-if-my-answer-is-wrong-i-cant-see-the-correct-answer-and-my-answer


src/components/cards.js
```diff
@@ -9,6 +9,8 @@ import withVibration from '../containers/with-vibration';
 import type {WithVibrationProps} from '../containers/with-vibration';
 
 export type Card = {|
+  answers?: Array<string>,
```
VincentCATILLON: Should use Answer type already defined

src/components/cards.js
```diff
@@ -9,6 +9,8 @@ import withVibration from '../containers/with-vibration';
...
+  userAnswers?: Array<string>,
```
VincentCATILLON: Same for this

PR #439, opened 2019-11-18
Use modal for selects

VincentCATILLON:
## Detailed purpose of the PR

This PR uses modal for select pickers

### Result

**Before**

Kapture 2019-11-19 at 10 56 35

**After**

Kapture 2019-11-19 at 10 52 16

### Testing strategy


jest.config.unit.js
```diff
@@ -33,13 +33,13 @@ module.exports = {
     },
     'src/containers': {
       // @todo increase this value
-      branches: 38,
+      branches: 41,
```
Volubyl: 👏

src/__fixtures__/store.test.js
```diff
@@ -253,9 +253,10 @@ describe('storeFixture', () => {
 
     const expectedResult: StoreState = {
       data: data,
-      error: {
+      errors: {
```
Volubyl: why plural ?

src/components/modal-select.js
```diff
@@ -0,0 +1,77 @@
...
+
+import * as React from 'react';
+import {View, FlatList, StyleSheet} from 'react-native';
+import type {Choice} from '@coorpacademy/progression-engine';
+
+import theme from '../modules/theme';
+import Modal from './modal';
+import ModalSelectItem from './modal-select-item';
+
+type ChoiceValue = $NonMaybeType<$PropertyType<Choice, 'value'>>;
```
Volubyl: flow 3th dan - black belt

src/components/modal.js
```diff
@@ -0,0 +1,100 @@
...
+  headerBackgroundColor,
+  iconBackgroundColor,
+  renderIcon,
+  contentStyle,
+  onClose,
+  testID
+}: Props) => (
+  <View style={styles.container} testID={testID}>
+    <View style={[styles.header, {backgroundColor: headerBackgroundColor}]}>
+      <HeaderBackButton
```
Volubyl: question why `BackButton` instead of `CloseButton` ?

src/redux/actions/progressions/create-next-progression.js
```diff
@@ -4,7 +4,7 @@ import {selectProgression, fetchBestProgression} from '@coorpacademy/player-stor
 import type {LevelAPI, ChapterAPI} from '@coorpacademy/player-services';
 import type {Progression} from '@coorpacademy/progression-engine';
 
-import type {StoreAction, ErrorAction} from '../../_types';
+import type {StoreAction, StoreErrorAction} from '../../_types';
```
Volubyl: does not understand the naming here is there any other kind of `ErrorAction` not related to the store ?

src/components/modal-select-item.js
```diff
@@ -0,0 +1,45 @@
...
+
+type Props = {|
+  isSelected?: boolean,
+  children: string,
+  onPress: () => void,
+  testID?: string
+|};
+
+const ModalSelectItem = ({isSelected, children, onPress, testID}: Props) => (
+  <BrandThemeContext.Consumer>
```
Fausto95: `useContext` here please

src/__fixtures__/store.test.js
VincentCATILLON: to have the same name everywhere (actions, reducers, etc), because having `error` and `errors` somestimes is boring

src/components/modal.js
VincentCATILLON: This component already exists in the project, this is the header back button, which can have 3 types:
- home
- close
- back

src/redux/actions/progressions/create-next-progression.js
VincentCATILLON: This type was introduced and confusing with the other `ErrorAction` (in `ui/errors`), and so, to avoid having 2 identical names, we changed it

PR #449, opened 2019-11-26
Fix/sync progression (403 & 409)

Fausto95:
### What it does and how?
It'll retry posting the progression whenever we get an 403(forbidden) error only, and it also avoids retrying when an 409(conflicting resource) error is thrown and consider the progression as synced.   

- [x] I added tests


src/redux/actions/progressions/synchronize.js
```diff
@@ -46,14 +46,12 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
...
-      // $FlowFixMe here the progression will always have an id cause we check before calling the function
-      synchronizedProgressionsIds.push(progression._id);
     } catch (error) {
-      if (error instanceof ForbiddenError) throw new Error(error);
-      if (numberOfRetries > 0) {
+      if (error instanceof ForbiddenError && numberOfRetries > 0) {
         await delay(2000);
         return syncProgression(progression, numberOfRetries - 1);
       }
+      if (error instanceof ConflictError) return Promise.resolve();
```
Volubyl: nope, here you should only `return`. No need of `Promise.resolve()`

Fausto95:
```suggestion
      if (error instanceof ConflictError) return;
```

src/models/error.test.js
```diff
@@ -15,4 +15,28 @@ describe('Error', () => {
       expect(error.stack).toBeDefined();
     });
   });
+  describe('NoContentFoundError', () => {
```
VincentCATILLON: Nice lead

PR #452, opened 2019-12-02
User-Agent

Fausto95:
Capture d’écran 2019-12-02 à 11 11 00

This PR adds specific User-Agent for easy track of http calls made using the app

- [x] Tested(incl. Datadog)


Fausto95:
Covered by #453

PR #453, opened 2019-12-02
Adding User-Agent for mobile

Fausto95:
Capture d’écran 2019-12-02 à 11 11 00

This PR adds specific User-Agent for easy track of http calls made using the app

- [x] Tested(incl. Datadog)


jest-setup.js
```diff
@@ -41,6 +41,12 @@ NativeModules.ReactLocalization = {
   language: 'en-US'
 };
 
+// react-native Platform
+NativeModules.PlatformConstants = {
```
Fausto95: Plaform.Version cannot me overwritten, that's why I'm mocking the version using NativeModules.PlatformConstants which is called by the `Version` getter method on the Platform module

src/components/version.js
```diff
@@ -21,7 +21,7 @@ const styles = StyleSheet.create({
 
 const Version = ({style}: Props) => (
   <Text style={[styles.version, style]}>
-    {translations.formatString('{0}: {1}', 'Version', `${version.tag}.${version.commit}`)}
+    {translations.formatString('{0}: {1}', 'Version', `v${version.tag}.${version.commit}`)}
```
Fausto95: I'm adding the `v` prefix here `cause we don't really need it in the `version.json` file nor in User-Agent string

PR #454, opened 2019-12-03
Fix/User-Agent

Fausto95:
Generate a real and parsable User-Agent

Capture d’écran 2019-12-03 à 14 10 18

Capture d’écran 2019-12-03 à 14 10 31


src/modules/fetch.js
```diff
@@ -2,17 +2,29 @@
...
+import DeviceInfo from 'react-native-device-info';
 
 import get from 'lodash/fp/get';
 import {ForbiddenError} from '../models/error';
 import version from './version';
 
 export const ERROR_MESSAGE = 'PLATFORM_DISABLED';
 
+const getUserAgent = async () => {
+  const deviceBrand = await DeviceInfo.getBrand();
```
Fausto95: Should Probably `Promise.all` these calls

src/modules/fetch.js
```diff
@@ -2,17 +2,29 @@
...
+  const deviceModel = await DeviceInfo.getModel();
+  const deviceVersion = await DeviceInfo.getSystemVersion();
+  const appVersion = `${version.tag}`.slice(1);
```
Fausto95: version.tag commes with a `v` prefix, so here I remote it since it breaks the version parsing

PR #455, opened 2019-12-03
migration script

Fausto95:
# Migration script


src/migrations/index.js
```diff
@@ -0,0 +1,55 @@
+import services from '../services';
+import {
+  checkIfNeedMigration,
+  getLatestMigrationVersion,
+  incrementAsyncStorageVersion
+} from './utils';
+
+const migrations = {
+  1: {
```
VincentCATILLON: should be a function

src/migrations/index.js
```diff
@@ -0,0 +1,55 @@
+import services from '../services';
+import {
```
VincentCATILLON: can you do this on `modules/migration` and test it so ?

src/migrations/index.js
```diff
@@ -0,0 +1,55 @@
...
+    progressions
+  };
+};
+
+const runMigrations = async () => {
+  const shouldMigrate = await checkIfNeedMigration();
+  if (shouldMigrate) {
+    try {
+      const latestMigrationVersion = getLatestMigrationVersion();
+      const data = await getAllAsyncStorageData();
```
VincentCATILLON: we don't need it

src/migrations/index.js
```diff
@@ -0,0 +1,55 @@
...
+      const migrateFns = Object.values(migrations[latestMigrationVersion]).map(async fn => {
+        await fn(data);
+      });
+      await Promise.all(migrateFns);
+      await incrementAsyncStorageVersion();
```
VincentCATILLON: the responsibility of this function is to check if we got migrations to run depending on our current version.

ie: version 1 in async storage, we got migrations 2 and 3. so, the aim of this function is to orchestrate the migration chaining one migration after one another (and persisting version after each migration: `1 -> 2` and `2 -> 3`) to avoid running twice if an error occurs in the 3rd migration

src/migrations/utils/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const getAsyncStorageVersion = async () => {
+  const version = await AsyncStorage.getItem(ASYNC_STORAGE_VERSION_KEY);
+  return version || 1;
+};
+
+export const incrementAsyncStorageVersion = async () => {
+  const currentVersion = await getAsyncStorageVersion();
+  await AsyncStorage.setItem(ASYNC_STORAGE_VERSION_KEY, currentVersion + 1);
+};
```
VincentCATILLON: I'd prefer `setAsyncStorageVersion` to be correlant with `getAsyncStorageVersion`

src/migrations/utils/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+export const getLatestMigrationVersion = migrations => {
+  const migrationVersions = Object.keys(migrations).sort((a, b) => a - b);
+  const latest = migrationVersions[migrationVersions.length - 1];
```
VincentCATILLON: looks dangerous

src/migrations/utils/index.js
```diff
@@ -0,0 +1,26 @@
+import AsyncStorage from '@react-native-community/async-storage';
+
+export const ASYNC_STORAGE_VERSION_KEY = 'async_storage_version';
+
+export const getAsyncStorageVersion = async () => {
+  const version = await AsyncStorage.getItem(ASYNC_STORAGE_VERSION_KEY);
```
VincentCATILLON: you have to try catch this, or the app will crash every time a user launch the app, without any version in the async storage

src/migrations/utils/index.js
```diff
@@ -0,0 +1,26 @@
...
+  const migrationVersions = Object.keys(migrations).sort((a, b) => a - b);
+  const latest = migrationVersions[migrationVersions.length - 1];
+  return latest;
+};
+
+export const checkIfNeedMigration = async () => {
+  const currentAsyncStorageVersion = await getAsyncStorageVersion();
+  const latestMigrationVersion = getLatestMigrationVersion();
+  if (latestMigrationVersion > currentAsyncStorageVersion) return true;
+  return false;
```
VincentCATILLON: `return latestMigrationVersion > currentAsyncStorageVersion;`

but IMO this thing should be in the migration orchestrator function

VincentCATILLON: edit: you could have `getNextMigrations` for example that is the list of migrations you have to run, and you can map over it, setting the version in async storage once each one has been finished

godu:
Il manque l'ordre. Les scripts doivent être lancer dans un certain ordre.
Si on passe de la v1 à la v3. Il faut lancer le script v2 avant le script v3.

godu:
u peux virer checkIfNeedMigration Si le reste de ton code est correct, la liste des scripts à lancer sera vide, et ça reviendra au même.

godu:
Par contre, la version du storage, c'est pas ouf. Autant prendre la version de l'app. On se fera moins chier.

src/migrations/types.js
```diff
@@ -0,0 +1,20 @@
+// @flow
+import type {Progression} from '@coorpacademy/progression-engine';
+import type {Resource} from '../layer/data/_types';
+
+type Content = Resource | Progression | string;
+
+export type Data = {
```
Fausto95: `The content here must reflect all(at least the most import ones) the data types that we have in the Async Storage`

src/migrations/tests/index.test.js
```diff
@@ -0,0 +1,234 @@
...
+      }
+    });
+
+    const asyncStoragedata = {
+      token: {
+        data: 'someFakeToken',
+        saveData: newData => AsyncStorage.setItem('@@token', newData)
+      },
+      progressions: {
+        data: [goodProgression],
```
Fausto95: The idea here is to provide the data and the associated function that'll save the data in the async stoorage, this makes it easier becauser the script function will only receive the data and save it directly.

src/migrations/tests/index.test.js
```diff
@@ -0,0 +1,234 @@
...
+    const AsyncStorage = require('@react-native-community/async-storage');
+    const migrations = {
+      '3': {
+        token: async ({token: {data, saveData}}) => {
+          // $FlowFixMe no data coertion.
+          const newToken = `${data}_@coorpacademy`;
+          await saveData(newToken);
+        }
+      },
+      '4': {
```
Fausto95: Also keys are passed as strings here because flow yeea

src/migrations/index.js
```diff
@@ -0,0 +1,48 @@
...
+  const allASKeys = await asynStorage.getAllKeys();
+
+  return Promise.mapSeries(allASKeys, async key => {
+    if (predicate(key)) {
+      const value = await asynStorage.getItem(key);
+      const nextValue = await migration(value, key);
+      const currentASVersion = (await asynStorage.getItem(asyncStorageVersionKey)) || 1;
+      if (value !== nextValue) {
+        await asynStorage.setItem(key, nextValue);
+        await asynStorage.setItem(asyncStorageVersionKey, currentASVersion + 1);
```
godu: Toute la gestion du `asyncStorageVersionKey` doit être faite dans `applyMigrations`

src/migrations/index.js
```diff
@@ -0,0 +1,45 @@
...
+  const currentASVersion = (await asyncStorage.getItem(asyncStorageVersionKey)) || 1;
+  return _.pipe(
+    _.toPairs,
+    _.map(_.update('0', _.parseInt(10))),
+    _.filter(([version]) => version > currentASVersion),
+    _.sortBy('0'),
+    vm =>
+      Promise.mapSeries(vm, async ([version, [predicate, migration]]) => {
+        await applyMigration(predicate, migration)(asyncStorage);
+        await asyncStorage.setItem(asyncStorageVersionKey, currentASVersion + 1);
```
godu:
```suggestion
        await asyncStorage.setItem(asyncStorageVersionKey, version);
```
C'est pas forcement currentASVersion + 1. Surtout si tu passes de la v1 à la v3, `currentASVersion + 1 === 2`  et non `3`.

jest.config.unit.js
```diff
@@ -89,6 +89,12 @@ module.exports = {
       function: 100,
       lines: 100,
       statements: 100
+    },
+    'src/migrations': {
+      branches: 100,
+      function: 100,
+      lines: 100,
+      statements: 100
```
VincentCATILLON: add the folder into `collectCoverageFrom` property too

PR #461, opened 2019-12-13
Catch errors

Fausto95:
Adding all awaited calls inside a try/catch block


VincentCATILLON:
Can you explain the aim ?

PR #462, opened 2019-12-19
Fix subtitle statement and author without label

VincentCATILLON:
## Detailed purpose of the PR

This PR fixes author without label and card without author errors

### Testing strategy


src/components/catalog-item-footer.js
```diff
@@ -131,7 +131,12 @@ const CatalogItemFooter = ({item, testID, size}: Props) => {
   const iconCertifiedSize = subtitleFontSize * 1.1;
 
   const author = getAuthor(item);
-  const subtitle = item.authors && item.authors.map(({label}) => label).join(', ');
+  const subtitle =
+    item.authors &&
+    item.authors
+      .filter(({label}) => label)
```
VincentCATILLON: to be sure that we only keep author with label

src/components/catalog-item-footer.js
```diff
@@ -165,7 +170,7 @@ const CatalogItemFooter = ({item, testID, size}: Props) => {
       >
         {item.title}
       </Text>
-      {subtitle && (
+      {subtitle ? (
```
VincentCATILLON: to avoid string printed outside of `Text` tag

PR #464, opened 2019-12-28
🖤 🌚Dark Mode  🌚🖤

Fausto95:
## Detailed purpose of the PR

This PR adds a new 🖤 🌚Dark Mode  🌚🖤 theming to the whole app.

There's no functionality breaking changes, only colors.

While you're here, the best scenario to test this PR is to have both Light & Dark apps running in different simulators(you should also consider having the Zeplin files to quick check the design choices) & compare.

## Some of the screens:

<div style="display:flex; flex-direction:row;flex-wrap:wrap">
DarkMode

DarkMode

DarkMode

DarkMode
</div>

#### Additional libs:

`react-native-appearance@0.3.0`

PS: The React-Native team is working on something similar that'll be built-in in `react-native` itself and will probably ship with the stable `v0.62`, I'll then refactor this that time.

# STILL TODO

- [ ] Fix connection screens button colors
- [ ] Fix StatusBar color
- [ ] Fix Button when finish a level on error
- [ ] Fix Correction and Level End Status Banner(maybe) 
- [ ] Iterate over some UIs
- [ ] Better Select Modal Colors(Design) 

### Testing strategy


VincentCATILLON:
👀

jest-setup.js
```diff
@@ -14,6 +14,11 @@ ScrollView.propTypes = {
   decelerationRate: () => {}
 };
 
+// React-Native Appearance
+jest.mock('react-native-appearance', () => ({
+  useColorScheme: () => 'dark'
```
VincentCATILLON: not sure of this

src/components/authentication.js
```diff
@@ -109,6 +109,7 @@ const Authentication = ({
           <Button
             isInverted
             isTextSecondary
+            isDarkModeActivated={false}
```
VincentCATILLON: Enabled please

jest-setup.js
Fausto95: Yeah, I'll make it light by default

PR #467, opened 2020-01-22
Locked content

VincentCATILLON:
## Detailed purpose of the PR

This PR adds support for locked content and triggers a specific error code when posting locked content.

[Diff without snapshots (+372 / -110)](https://github.com/coorpacademy/mobile/compare/master...ebde98b505ea4464d01df74817bf769cbac361d1)

### Result

Kapture 2020-01-23 at 12 08 34

### Testing strategy


src/redux/actions/progressions/synchronize.js
```diff
@@ -47,11 +47,14 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
...
-        await delay(2000);
-        return syncProgression(progression, numberOfRetries - 1);
-      }
       if (error instanceof ConflictError) return;
+      if (error instanceof ForbiddenError) {
+        if (numberOfRetries > 0) {
+          await delay(2000);
+          return syncProgression(progression, numberOfRetries - 1);
+        }
+        throw new ForbiddenProgressionError(error.message);
```
VincentCATILLON: This thing to avoid the middleware to catch this error and to see "Platform disabled" modal

src/components/catalog-item.js
```diff
@@ -76,46 +78,60 @@ class CatalogItem extends React.PureComponent<Props> {
       section
     };
 
+    const isLocked = item && item.accessible === false;
```
VincentCATILLON: this thing is to avoid having locked content whereas we don't have the property set on the card

PR #474, opened 2020-02-20
Fix card display in level end and explicit deck cards names

VincentCATILLON:
## Detailed purpose of the PR

This PR fixes a bad card integration due to `Card` component refacto, so as everytime we are editing this component we got a regression, it also splits the `Card` component and explicits everything used in deck swiper.

##### Before

Capture d’écran 2020-02-20 à 16 49 21

##### After

Capture d’écran 2020-02-20 à 16 08 34
Capture d’écran 2020-02-20 à 16 08 13
Capture d’écran 2020-02-20 à 16 08 48

### Testing strategy


Fausto95:
Tested on real Android device ✅

VincentCATILLON:
Tested on real iPhone device ✅

Fausto95:
image

One of the Android device may has this bug

PR #476, opened 2020-02-25
Fix 409 Error

Fausto95:
This pull request fixes 409 errors.

### Logic:

The current master behavior is that we treat the progression has been synced after all progression being synced.

This PR treats each progression as synced or pending at the end of its sync step

Before

image

After 

image

datadog: https://bit.ly/2HVMl9w


src/layer/data/progressions.js
```diff
@@ -85,10 +86,19 @@ const getSynchronizedProgressionIds = async (): Promise<Array<string>> => {
   return _synchronizedProgressions ? JSON.parse(_synchronizedProgressions) : [];
 };
 
+const getPendingProgressionId = async (): Promise<string> => {
+  const pendingProgressionId: string = await AsyncStorage.getItem(PENDING_PROGRESSION);
```
Fausto95: Since this was supposed to be a queue that'd hold a single element, I preferred  using a string instead that way is easier to get the current element

src/layer/data/progressions.js
```diff
@@ -124,6 +134,31 @@ const synchronize = async (
   return;
 };
 
+const progressionHasBeenAlreadyPosted = async (
```
Fausto95: This will make an API call to check if the given progressionId is known(been posted before) and return a boolean

src/layer/data/progressions.js
```diff
@@ -124,6 +134,31 @@ const synchronize = async (
...
+  const response = await fetch(`${host}/api/v2/progressions/${progressionId}`, {
+    method: 'GET',
+    headers: {
+      'X-Requested-With': 'XMLHttpRequest',
+      'Content-Type': 'application/json',
+      Authorization: token
+    }
+  });
+
+  if (response.status === 404) return false;
```
Fausto95: Whenever we pass a progressionId that did not been posted before, the API will return a 404 error, so in that case we want to return false so that the progression can be then synchronized

src/layer/data/progressions.js
```diff
@@ -124,6 +134,31 @@ const synchronize = async (
...
+  if (response.status >= 400) throw new Error(response.statusText);
+  if (response.status === 200) {
+    const progression: Progression = await response.json();
+    const hasBeenPosted: boolean = !!(progression && progression._id);
+    return hasBeenPosted;
+  }
+  return false;
```
Fausto95: Worts case, we just return false

src/redux/actions/progressions/synchronize.js
```diff
@@ -42,6 +42,7 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
 
   const progressions = await services.Progressions.getAll();
   const synchronizedProgressionsIds: Array<string> = await services.Progressions.getSynchronizedProgressionIds();
+  const pendingProgressionId: string = await services.Progressions.getPendingProgressionId();
```
Fausto95: This should probably be done right in the async pMap function

src/redux/actions/progressions/synchronize.js
```diff
@@ -67,11 +68,30 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
...
+          synchronizedProgressionsIds.push(pendingProgressionId);
+          await services.Progressions.updateSynchronizedProgressionIds(synchronizedProgressionsIds);
+          await services.Progressions.updatePendingProgressionId('');
+          return dispatch({
+            type: SYNCHRONIZE_SUCCESS,
+            meta: {id: pendingProgressionId}
+          });
+        }
         try {
+          await services.Progressions.updatePendingProgressionId(_id);
```
Fausto95: We add the progression's ID  to pending, if everything goes as planned, we then remove it after the sync

src/redux/actions/progressions/synchronize.js
```diff
@@ -67,11 +68,30 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
...
       if (_id) {
+        if (pendingProgressionId) {
+          progressionHasBeenPostedAlready = await services.Progressions.progressionHasBeenAlreadyPosted(
+            token,
+            brand.host,
+            pendingProgressionId
+          );
+        }
+        if (progressionHasBeenPostedAlready) {
+          synchronizedProgressionsIds.push(pendingProgressionId);
```
Fausto95: If the progression is already pending and has been posted in the API, we then treat it as sync and move on, otherwise we do sync it

src/layer/data/index.js
```diff
@@ -7,11 +7,14 @@ import {
   findById as findProgressionById,
   getAll as getAllProgressions,
   getSynchronizedProgressionIds,
+  getPendingProgressionId,
+  progressionHasBeenAlreadyPosted,
```
VincentCATILLON: ?

VincentCATILLON: findRemoteById looks better

src/layer/data/progressions.js
```diff
@@ -16,6 +16,7 @@ import {ForbiddenError, ConflictError, NotAcceptableError} from '../../models/er
 import type {Record, Completion, HeroRecommendation} from './_types';
 
 export const SYNCHRONIZED_PROGRESSIONS = 'synchronized_progressions';
+export const PENDING_PROGRESSION = 'pending_progression';
```
VincentCATILLON: 👌

src/layer/data/progressions.js
```diff
@@ -85,10 +86,19 @@ const getSynchronizedProgressionIds = async (): Promise<Array<string>> => {
...
+  return pendingProgressionId || '';
+};
+
 const updateSynchronizedProgressionIds = (synchronizedProgressions: Array<string>): void => {
   AsyncStorage.setItem(SYNCHRONIZED_PROGRESSIONS, JSON.stringify(synchronizedProgressions));
 };
 
+const updatePendingProgressionId = (pendingProgressionId: string): void => {
+  AsyncStorage.setItem(PENDING_PROGRESSION, pendingProgressionId);
+};
```
VincentCATILLON: should return a promise (the `setItem` result directly)

src/layer/data/progressions.js
```diff
@@ -124,6 +134,31 @@ const synchronize = async (
...
+
+  if (response.status === 404) return false;
+  if (response.status >= 400) throw new Error(response.statusText);
+  if (response.status === 200) {
+    const progression: Progression = await response.json();
+    const hasBeenPosted: boolean = !!(progression && progression._id);
+    return hasBeenPosted;
+  }
+  return false;
+};
```
VincentCATILLON: imo this function should return the progression instead of a boolean, and you will be able to use it in multiple cases, in yours, you only need to know if the progression is undefined or not

src/redux/actions/progressions/synchronize.js
```diff
@@ -67,11 +68,30 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
     ),
     async (progression): Promise<Action | void> => {
       const {_id} = progression;
-
+      let didFindProgression: boolean = false;
```
VincentCATILLON: why using `let ` as it's not reassigned ?

src/redux/actions/progressions/synchronize.js
```diff
@@ -67,11 +68,30 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
...
+        }
+        if (didFindProgression) {
+          synchronizedProgressionsIds.push(pendingProgressionId);
+          await services.Progressions.updateSynchronizedProgressionIds(synchronizedProgressionsIds);
+          await services.Progressions.updatePendingProgressionId('');
+          return dispatch({
+            type: SYNCHRONIZE_SUCCESS,
+            meta: {id: pendingProgressionId}
+          });
+        }
```
VincentCATILLON: imo doing this is easier to understand:

```
const pendingProgressionId = await ...;
const remoteProgression = await services.Progressions.findRemoteProgressionById(...);
[...]
if (!remoteProgression) {
  await services.Progressions.updatePendingProgressionId(_id);
  await syncProgression(progression);
}
synchronizedProgressionsIds.push(_id);
await services.Progressions.updateSynchronizedProgressionIds(synchronizedProgressionsIds);
await services.Progressions.updatePendingProgressionId('');
```

src/redux/actions/progressions/synchronize.js
VincentCATILLON: 🔝 obsolete if you do the thing below

src/redux/actions/progressions/synchronize.js
```diff
@@ -68,25 +67,27 @@ export const synchronizeProgressions: StoreAction<Action> = async (dispatch, get
...
+          if (remotePogression && remotePogression._id) {
+            synchronizedProgressionsIds.push(pendingProgressionId);
+            await services.Progressions.updateSynchronizedProgressionIds(
+              synchronizedProgressionsIds
+            );
+            await services.Progressions.updatePendingProgressionId('');
+            return dispatch({
+              type: SYNCHRONIZE_SUCCESS,
+              meta: {id: pendingProgressionId}
+            });
```
VincentCATILLON: this thing is a logic duplication of https://github.com/CoorpAcademy/mobile/pull/476/commits/31ffb6e39b686c9f0733794c616260f466a18f28#diff-b6a4f15ef5101d7be99b8ac85ff9568fR96 and is not needed

PR #484, opened 2020-03-11
Fix not appearing text when searching

Fausto95:
# Before 

text-input

# After

text-input-after


src/components/search-input.js
```diff
@@ -27,7 +27,14 @@ const styles = StyleSheet.create({
   input: {
     flex: 1,
     color: theme.colors.black,
-    fontSize: theme.fontSize.large
+    fontSize: theme.fontSize.large,
+    // When we don't se any height, the text doesn't appear on Android
```
Fausto95:
```suggestion
    // When we don't set any height, the text doesn't appear on Android
```

PR #485, opened 2020-03-17
Turn Search UI into a screen

Fausto95:
Turns the current Search UI into a navigable screen :)

The current behavior just toggles the search ui and the catalog ui whenever we click in the magnifier search icon, which leads into some performance issues(lags)

By turning it into a screen, we do not have to deal with the catalog fetch cycle or nothing else, we just navigate to the search ui and do whatever we want right in there :)

The only breaking change user facing is the navigation, the rest of the work flow still the same :)

### [After](https://kickassstarter.s3.eu-west-2.amazonaws.com/screen-search.gif) (click to see gif)


silouone:
ah ouai pas de gif de preview  😱 c'est comme ça

jomaora:
je pense un peu comme @silouone ... Vu que la PR est demandé à reviewer à des non mobile, un peu de contexte serait bienvenu @Fausto95

src/containers/search.js
```diff
@@ -0,0 +1,103 @@
...
+export type Props = {|
+  ...ConnectedStateProps,
+  ...ConnectedDispatchProps,
+  ...ReactNavigation$ScreenPropsWithParams<Params>,
+  onCardPress: (item: DisciplineCard | ChapterCard) => void,
+  onBackPress: () => void
+|};
+
+export const SEARCH_DEBOUNCE_DURATION = 500;
+export const SEARCH_MIN_LENGTH = 2;
```
alanlanglois: You got these values in two different files (containers and components), wouldn't be better to have it to only one place?

Fausto95: Will do it!

PR #486, opened 2020-03-18
Skip fetch in case of e2e

VincentCATILLON:
## Detailed purpose of the PR

This PR throw errors when using fetch in e2e modes to avoid logs on datadog.

### Testing strategy


jest-setup.js
```diff
@@ -6,8 +6,8 @@ import mockAsyncStorage from '@react-native-community/async-storage/jest/async-s
 // AsyncStorage
 jest.mock('@react-native-community/async-storage', () => mockAsyncStorage);
 
-// global mocks
-global.fetch = jest.fn().mockImplementation(() => Promise.resolve());
```
Fausto95: We were just mocking the built-in `fetch`? 😂

VincentCATILLON: Yes 🤦‍♂

PR #487, opened 2020-03-18
Fix409

alanlanglois:
## Detailed purpose of the PR
Fixing 409 error when trying to update progression. 
We check if the remote progression exists no matter what, if it exists we skip the update, if not we update. 

### Testing strategy


Fausto95:
So you can also delete all the fonctions related to `pendingProgression`

alanlanglois:
@Fausto95  Yes you're right, I was not sure it was used somewhere else, done in my last commit

PR #490, opened 2020-03-19
TypeScript 🎉

Fausto95:
# I'm just having fun :) 

This is a WIP(just installed Typescript to see it working). I'll be creating sub-PRs for each file or each folder :)

I'll be adding more todos

# TODO 
- [x] Make sure the app launches
- [ ] Fix linting and type checking
- [x] Migrate each and every directory
- [ ] Install Prettier & TS Jest
- [x] Minimum compiler options for TS
- [ ] Fix common type errors
- [ ] Type Coorp libraries(fix path)
- [ ] Fix libraries types by installing `@types/$libname`(except for coorp ones)
- [ ] Fix most of JSX typing
- [x] Remove all `// @flow` comments
- [x] Replace all `// $FlowFixMe` comments with `// @ts-ignore`(FTM, I will be fixing typing issues in follow up PRs)
- [x] Make e2e tests work with TS(right now it launches the app, but fails on TS compiler React errors) 
- [x] Make unit tests work with TS(some of the unit tests are passing)
- [x] Make Jest snapshots tests work with TS(I can't make it work, I dunno how to)
- [ ] Use TS everywhere(inc. scripts)

# Testing

<small><b>For tracking reason, last update was March 29</b></small>


AdrieanKhisbe:
:scream:

tsconfig.json
```diff
@@ -0,0 +1,155 @@
...
+    // "inlineSources": true,                 /* Emit the source alongside the sourcemaps within a single file; requires '--inlineSourceMap' or '--sourceMap' to be set. */
+
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+  },
+  "exclude": [
+    "node_modules", "babel.config.js", "metro.config.js", "jest.config.js"
+  ]
+}
```
VincentCATILLON: just an idea 💡 maybe we can reuse TS config from other packages (lambdas for example) like we do with eslint to share config

package.json
```diff
@@ -165,10 +167,12 @@
     "opn-cli": "^4.0.0",
     "react-dom": "16.8.3",
     "react-native-firebase": "5.5.6",
+    "react-native-typescript-transformer": "^1.2.13",
```
Fausto95: Might not be needed

PR #491, opened 2020-03-19
Omniplayer (with modal)

VincentCATILLON:
## Detailed purpose of the PR

This PR adds omniplayer support to the app using quick win solution (webview) in a modal (like PDF)

### Testing strategy


VincentCATILLON:
should be closed because of multiple things changed (PDF, etc)

PR #492, opened 2020-03-19
Omniplayer video incrusted

VincentCATILLON:
## Detailed purpose of the PR

Same than https://github.com/CoorpAcademy/mobile/pull/491 but incrusted a video player (and with a preview loading), and so, without modal.

### Result

Capture d’écran 2020-03-19 à 17 26 48
Capture d’écran 2020-03-19 à 17 26 39

### Testing strategy


src/__fixtures__/lessons.js
```diff
@@ -2,23 +2,29 @@
...
+  videoId?: string
 }): Lesson => ({
   _id: ref,
   poster,
   description,
-  videoId: 'KovTu3zU',
+  videoId,
   mediaRef: 'med_jwp_Vy4JQKFhN',
-  mimeType: 'application/jwplayer',
+  // $FlowFixMe @todo change Lesson type in progression engine
```
VincentCATILLON: temporary thing that should be done once we wanna do things properly

src/layer/data/_const.js
```diff
@@ -62,14 +62,19 @@ export const RESTRICTED_RESOURCE_TYPE: {|
   k => RESTRICTED_RESOURCE_TYPE[k]
 ): Array<RestrictedResourceType>);
 
+// @todo change it in player services
+type VideoProvider = PlayerVideoProvider | 'omniPlayer';
```
VincentCATILLON: temporary thing that should be done once we wanna do things properly

src/components/resource.js
```diff
@@ -52,19 +82,38 @@ class Resource extends React.PureComponent<Props> {
...
       return null;
     }
 
+    const height = layout.width / (16 / 9);
+
     switch (type) {
       case RESOURCE_TYPE.VIDEO: {
+        const videoProvider = mimeType && getVideoProvider(mimeType);
+
+        if (!videoProvider) {
```
Fausto95: error prone

VincentCATILLON: @Fausto95 can you explain your feeling ?

Fausto95: We have a mimeType, so we do also have a provider.
This will break if we update coorp deps

VincentCATILLON: The mime type is always defined, but we are getting the video provider related to mime type, but in some cases (in this one), we don't have provider as well as `video/mp4` mime type

VincentCATILLON:
I'll do it in another PR as it's urgent

PR #496, opened 2020-03-24
manage events & cleanup

Fausto95:
## Detailed purpose of the PR

> :wave: Write here :wink:

### Testing strategy


src/containers/external-video-controlable.js
```diff
@@ -0,0 +1,133 @@
...
+    this.props.onPlay();
+  };
+
+  handleError = () =>
+    this.setState({
+      step: STEP.ERROR
+    });
+
+  handleExpand = async () => {
+    await this.props.toggleFullscreen(true);
```
Fausto95: Currently not being used, need to figure out how to make it work

Fausto95:
Need to clean up a bit more(maybe)
Need to handle full-screen expansion
Need to add tests and fix the broken ones
Maybe share styles between external-video & video, but this would be error prone

src/containers/external-video-controlable.js
alanlanglois: Don't think we need it as fullscreen is manage by the youtube extension

PR #507, opened 2020-04-07
Hide  Lives (Header Right) when adaptive content

alanlanglois:
## Detailed purpose of the PR

Hide Header Right containing lives when content is Adaptive. 
https://trello.com/c/SKEsHgUA/1679-do-not-display-lives-and-counter-everywhere-when-module-or-chapter-isconditional
before: 
image
after: 
image

### Testing strategy

 - checked on iOS Simulator


src/containers/header-slide-right.test.js
```diff
@@ -0,0 +1,72 @@
+// @flow
+import renderer from 'react-test-renderer';
+import * as React from 'react';
+import {Component as HeaderSlideRight} from './header-slide-right';
+
+describe('header-slide-right', () => {
```
VincentCATILLON, line 6: HeaderSlideRight please :)

src/containers/header-slide-right.test.js
```diff
@@ -0,0 +1,72 @@
...
+        isGodModeEnabled={isGodModeEnabled}
+        isFastSlideEnabled={isFastSlideEnabled}
+        isAdaptive={isAdaptive}
+        toggleGodMode={jest.fn()}
+        toggleFastSlide={jest.fn()}
+      />
+    );
+
+    const headerSlideRightComponent = component.root.find(el => el.props.onPress);
+    expect(headerSlideRightComponent.props).toBeDefined();
```
VincentCATILLON, line 27: i dont understand what this test is supposed to test ?

alanlanglois: it checks the component HeaderSlideRight exists and has been rendered

VincentCATILLON: Maybe checking that props are defined is confusing for someone reading this test.
imho you should only test that headerSlideRightComponent is defined :)

alanlanglois: This way? expect(headerSlideRightComponent).toBeDefined();

Fausto95:
I just figured out that the current implementation might no be the right way to go. That said, we might want to have the property `livesDisabled` set to false(in the engine configuration) whenever we want to consume an adaptive content.

alanlanglois:
I'm not sure I get it?

Fausto95:
Anyway, we still got the flash

Fausto95:
Fixed by: https://github.com/CoorpAcademy/mobile/pull/524

PR #508, opened 2020-04-07
Add BR support

Monia-LAOUFI:
https://trello.com/c/BIrMy2sz 
## Detailed purpose of the PR

> interpret \<br /> as a line break

### Testing strategy

### After

Capture d’écran 2020-04-07 à 21 32 05


VincentCATILLON:
lgtm but i worked on, so i can't validate 👍 please check ci and unit tests (snapshots and coverage)

VincentCATILLON:
should be tested functionally

src/components/question-template.js
```diff
@@ -1,14 +1,16 @@
 // @flow
 
 import * as React from 'react';
-import {View, StyleSheet} from 'react-native';
+import {StyleSheet, FlatList} from 'react-native';
+import {last, trim} from 'lodash/fp';
```
Fausto95, line 5:
```suggestion
import trim from 'lodash/fp/trim';
import last from 'lodash/fp/last';
```

Fausto95: For the record, the previous `import {last, trim} from 'lodash/fp';` imports the whole library and then extracts the functions that we want.
With this `import last from 'lodash/fp/last';` instead we only import the functions that we need.

Fausto95:
> should be tested functionally

Yea. And since I don't reckon exactly what this is about, I should be great too add the before picture too. 
@VincentCATILLON do you have a specific content to test on?

VincentCATILLON:
@Fausto95 you can reproduce with this content : https://onboarding.coorpacademy.com/discipline/dis_V1D-OfxiH

I added a before screen too

VincentCATILLON:
@Fausto95 I also added an e2e use case to reproduce this carriage return case in a template.

PR #509, opened 2020-04-08
Adaptive Support

Fausto95:
This PR aims to support adaptive courses to be consumed on mobile app

### Testing strategy


src/redux/actions/ui/search.js
```diff
@@ -10,14 +12,14 @@ export type Action =
...
-      payload: string
+      payload: string | QueryParams
     |}
   | {|
       type: '@@search/FETCH',
       payload: boolean
     |};
 
-export const edit = (payload: string): Action => ({
+export const edit = (payload: string | QueryParams): Action => ({
```
VincentCATILLON, line 22: hm, strange thing, why it's not static ?

PR #511, opened 2020-04-14
Adaptive - removing lives

alanlanglois:
[Trello](https://trello.com/c/SKEsHgUA/1679-do-not-display-lives-and-counter-everywhere-when-module-or-chapter-isconditional)
## Detailed purpose of the PR

Do not display lives( and counter) everywhere (when module or chapter isConditional)

Before: 
image

After:
image

### Testing strategy

- [ ] I ensured my code is tested: 
- [ ] I took care to mention breaking changes and extra libs if appropriate

PR #513, opened 2020-04-14
Adaptive flow navigation

Fausto95:
# Adaptive flow navigation

This PR aims to support the Adaptive Content's flow navigation, which means that for some content we end up show the correction screen and for other we just don't.

### Testing strategy


Fausto95:
This PR only works properly with https://github.com/CoorpAcademy/mobile/pull/511 's implementation.

Fausto95:
Ok, I'll be fixing the props tests tomorrow, I'm off

src/redux/actions/ui/navigate.js
```diff
@@ -0,0 +1,66 @@
...
+  // when content is adaptive and has not been finished
+  // navigate to context or next question
+  if (isAdaptive) {
+    if (!isContentFinished) {
+      return NavigationActions.navigate({routeName: hasContext ? 'Context' : 'Question'});
+    }
+
+    // when content is adaptive and has been finished
+    // navigate to level end
+    return NavigationActions.navigate({
```
Fausto95, line 37: We can either use `NavigationActions` and then `navigation.dispatach` or we can simply return an object here and use `navigation.navigate`.

Both ways work tho, so :)

src/redux/actions/ui/navigate.js
```diff
@@ -0,0 +1,66 @@
...
+import type {StoreState} from '../../store';
+import {
+  isCorrect as _isCorrect,
+  getContext,
+  getLivesCount,
+  isContentFinished as _isContentFinished,
+  hasSuccessfullyFinished
+} from '../../utils/state-extract';
+
+export const getNextScreen = () => (dispatch: Dispatch, getState: GetState, options: Options) => {
```
Fausto95, line 17: This can /cannot be an action, it could be a state-extracter, but it makes more sense to be an action :)

src/redux/actions/ui/navigate.js
```diff
@@ -0,0 +1,66 @@
...
+  const slide = getPreviousSlide(state);
+  const lives = getLivesCount(state);
+  const isContentFinished = _isContentFinished(state);
+  const isContentFinishedSuccessfully = lives === undefined || lives > 0;
+  const isCorrect = _isCorrect(state);
+  const hasContext = getContext(state) !== undefined;
+  const currentProgressionId = getCurrentProgressionId(state);
+  const isAdaptive = typeof isCorrect !== 'boolean';
+
+  // when content is adaptive and has not been finished
```
Fausto95, line 28: Might drop the comments, just used them to help me figure out the way this should be implemented.

src/redux/utils/state-extract.js
```diff
@@ -184,3 +186,33 @@ export const getFocusedSelect = (state: StoreState): SelectState => state.select
...
+  const isFinished = _isExitNode || isOfferingExtraLife;
+  return isFinished;
+};
+
+export const hasSuccessfullyFinished = (state: StoreState): boolean => {
+  const nextContent = getStepContent(state);
+  if (!nextContent) {
+    return false;
+  }
+  if (isExitNode(state)) {
```
Fausto95, line 214: Just checking it twice here, cuz you know, If you check twice, no bugs tho ;)

src/screens/question.js
```diff
@@ -89,15 +90,10 @@ class QuestionScreen extends React.PureComponent<Props> {
...
-    const {slideId} = this.props;
-
-    const params: CorrectionParams = {
-      slideId
-    };
-
-    this.props.validateAnswer();
-    this.props.navigation.navigate('Correction', params);
+  handleButtonPress = async () => {
+    await this.props.validateAnswer();
```
Fausto95, line 94: Actually, dunno if I should put this lines in an async function then call it here.
Feedbacks wanted

src/redux/actions/ui/navigate.test.js
```diff
@@ -0,0 +1,296 @@
...
+    expect(result).toEqual({
+      type: 'Navigation/NAVIGATE',
+      routeName: 'LevelEnd',
+      params: {
+        isCorrect: true,
+        progressionId: 'progression1'
+      }
+    });
+  });
+});
```
VincentCATILLON, line 296: why don't we have navigation service (like router, or anything else) and mutualize this ?
btw, doing this in redux is useless as it doesn't change any state, it only call react navigation state. should be done in question screen so

PR #514, opened 2020-04-16
Adap flow nav extr

Fausto95:
# Refactor Action to State Extractor

### Testing strategy


src/types.js
```diff
@@ -137,3 +137,12 @@ export type LoggerProperties = {
...
 export type AuthenticationType = 'qr-code' | 'magic-link' | 'demo' | 'reconnection';
+
+export type NextScreenParams = {|
+  screenName: 'LevelEnd' | 'Question' | 'Context' | 'Correction',
+  params?: {
+    isCorrect?: boolean,
+    progressionId?: string,
+    slideId?: string
+  }
+|};
```
VincentCATILLON, line 148: arbitrary type which is not linked to screen params types, so, if you change the screen params type, nobody will tell you it's wrong

src/redux/utils/state-extract.js
```diff
@@ -216,3 +224,53 @@ export const hasSuccessfullyFinished = (state: StoreState): boolean => {
   }
   return false;
 };
+
+export const getNextScreen = (state: StoreState): NextScreenParams => {
```
VincentCATILLON, line 228: we are in state extract area, we should not deal with screens or screens params, we should extract data from state, and this logic should be done in the right place (in screens, or navigation area)

src/screens/question.js
```diff
@@ -89,9 +91,43 @@ class QuestionScreen extends React.PureComponent<Props> {
...
+      isContentFinished,
+      progressionId,
+      slideId
+    } = getContentCorrectionInfo(state);
+
+    if (isAdaptive) {
+      if (!isContentFinished) {
+        return navigate(hasContext ? 'Context' : 'Question');
+      }
+      const levelEndParams = {
```
VincentCATILLON, line 111: typing please

src/types.js
```diff
@@ -138,11 +138,11 @@ export type LoggerProperties = {
...
 export type AuthenticationType = 'qr-code' | 'magic-link' | 'demo' | 'reconnection';
 
-export type NextScreenParams = {|
-  screenName: 'LevelEnd' | 'Question' | 'Context' | 'Correction',
-  params?: {
-    isCorrect?: boolean,
-    progressionId?: string,
-    slideId?: string
-  }
+export type ContentCorrectionInfo = {|
```
VincentCATILLON, line 141: should be defined in state extract file as it's not reused elsewhere :)

PR #516, opened 2020-04-16
Link to search

alanlanglois:
## Detailed purpose of the PR

Go to search screen with given search params when clicking a link in the end popin for Adaptive
[trello](https://trello.com/c/lD4ZJNr2/1682-go-to-search-screen-with-given-search-params-when-clicking-a-link-in-the-end-popin-for-adaptive)

### Testing strategy


src/modules/uri.js
```diff
@@ -40,7 +41,30 @@ export const buildUrlQueryParams = (params: QueryParams) =>
...
+    'theme',
+    'author',
+    'type',
+    'authorType',
+    'lpath',
+    'sort',
+    'goal',
+    'search',
+    'playlist'
+  ];
```
VincentCATILLON, line 60: this kind of data, is related to search responsibility, this function should only return query params, and another function should filter properties

src/modules/uri.test.js
```diff
@@ -31,4 +31,19 @@ describe('Uri', () => {
       expect(buildUrlQueryParams(params)).toBe(expectedResult);
     });
   });
+
+  describe('get query params', () => {
```
VincentCATILLON, line 35: camel case to retrieve it easily please

src/redux/actions/ui/search.test.js
```diff
@@ -18,6 +18,20 @@ describe('search', () => {
     });
   });
 
+  describe('edit query params', () => {
```
VincentCATILLON, line 21: camel case

src/redux/actions/ui/search.js
```diff
@@ -22,6 +29,11 @@ export const edit = (payload: string): Action => ({
   payload
 });
 
+export const editQueryPrams = (payload: QueryParams): Action => ({
+  type: EDIT_QUERY_PARAMS,
+  payload
+});
```
VincentCATILLON, line 35: why do we need this action ? as it should be called and dispatched in payload of search

src/screens/level-end.js
```diff
@@ -100,6 +101,11 @@ class LevelEndScreen extends React.PureComponent<Props, State> {
     this.props.navigation.navigate('PdfModal', pdfParams);
   };
 
+  handleFeedbackLinkPress = (url: string) => {
+    const params = getQueryParams(url);
```
VincentCATILLON, line 105: type

src/screens/search.js
```diff
@@ -11,7 +11,7 @@ import {HEADER_BACKGROUND_COLOR} from '../navigator/navigation-options';
 import Search from '../containers/search';
 
 export type Params = {|
-  url: string
+  url: string // is this in order to manage param in search?
```
VincentCATILLON, line 14: ?

src/screens/level-end.js
```diff
@@ -100,6 +101,11 @@ class LevelEndScreen extends React.PureComponent<Props, State> {
...
+    this.props.navigation.navigate('Search', params);
```
VincentCATILLON, line 106: params should be typed as `SearchParams`

Fausto95: getQueryParams has return type, so `const params = ` here is type inferenced

src/modules/uri.js
Fausto95: I mean, either ways work here..

src/screens/level-end.js
VincentCATILLON: Seen with you, react navigation doesn't make the link between the params and the screen params we defined

src/redux/reducers/ui/search.js
```diff
@@ -17,15 +17,15 @@ export const initialState: State = {
 const reducer = (state: State = initialState, action: Action): State => {
   switch (action.type) {
     case EDIT: {
+      if (typeof action.payload === 'string') {
```
Fausto95, line 20: Know someday I'll blame myself for writing code like this.
Should probably consider payload as object always and have something like
```
payload: {
 value: 'mySearchValue',
 params: { skill: 'someSkill', ...}
}

File: src/utils/search.js [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+    'playlist'
+  ];
+  // $FlowFixMe
+  return allowedParams.reduce((acc, allowedParam) => {
+    if (allowedParam in params && params[allowedParam] !== undefined) {
+      acc[allowedParam] = params[allowedParam];
+      return acc;
+    }
+    return acc;
+  }, {});
```
VincentCATILLON, line 37: I know that `lodash` flow coverage is not as good as we could use, but, if this function has unit test, we could consider using https://gist.github.com/jfmengels/6b973b69c491375117dc#_pickprops-object

PR #524, opened 2020-04-27
Fixing glitch lives / progression bar between questions

alanlanglois:
## Detailed purpose of the PR

Aims to fix a quick glitch visible when moving from one question to another. 

### Result

Kapture 2020-04-28 at 17 53 48

### Testing strategy

- [x] I ensured my code is tested: unit test
- [x] I took care to mention breaking changes and extra libs if appropriate: `@coorpacademy/player-store@4.1.16`


src/containers/progression.js
```diff
@@ -26,11 +26,14 @@ const Progression = ({isHidden, current, count}: Props) => {
   return <ProgressionComponent current={current} count={count} />;
 };
 
-export const mapStateToProps = (state: StoreState): ConnectedStateProps => ({
-  isHidden: isContentAdaptive(state) || isVideoFullScreen(state),
-  current: getCurrentStep(state),
-  count: getNbSlides(state)
-});
+export const mapStateToProps = (state: StoreState): ConnectedStateProps => {
+  const {hide} = getLives(state);
```
VincentCATILLON, line 30: Progression is not related to lives, you can have progression but no lives, but both should be hidden in adaptive you're right

PR #530, opened 2020-05-06
Don't reorder correction cards on the resource viewed condition

Fausto95:
Correction cards  were being reordered when a resource was marked as viewed thus the jump while playing a video.

### Testing strategy


alanlanglois:
With this fix we loose a nice feature: give a correction card not seen by the user as the first card on the list. It might be tricky to keep it but I think it would be better. 
We should avoid the video to send "isResourceViewed" when played in correction screen.

Fausto95:
Closing in favor of #538

PR #532, opened 2020-05-12
Quick fix for sound in correction screen

alanlanglois:
## Detailed purpose of the PR

Because the way of displaying correction screen changed since the introduction of Adaptive courses, this is a quick fix to add sound and vibration feedback again in Correction Screen. 

IMO, we'll need to investigate more with Faust to find a better way to deal with this.

### Testing strategy

- [x] I ensured my code is tested: test-unit / lint / flow / emulators


alanlanglois:
See this comment, IMO it explains some of the bugs from the Adaptive support: https://github.com/CoorpAcademy/mobile/commit/fbd236d55b48909fbc890cdf32ae7004c5bde380#r39121588

PR #535, opened 2020-05-19
testing react-native\'s back-handler

Fausto95:
Adding tests for RN's `BackHandler` module

### Testing strategy


src/modules/back-handler.js
```diff
@@ -0,0 +1,29 @@
...
+    BackHandlerEmitterEvents[eventName] = callback;
+    delete BackHandlerEmitterEvents[eventName];
+  }),
+  exitApp: jest.fn(),
+  fireEvent(eventName) {
+    BackHandlerEmitterEvents[eventName]();
+  }
+};
+
+// eslint-disable-next-line fp/no-proxy
```
Fausto95, line 20: lame :)

src/modules/back-handler.js
```diff
@@ -0,0 +1,29 @@
...
+export const BackHandler = new Proxy(BaseBackHandler, {
+  get(target, prop) {
+    /* istanbul ignore next */
```
Fausto95, line 23: Dont @ me, this is needed since we don't really need to test the behavior of the original BackHandler

src/modules/back-handler.js
```diff
@@ -0,0 +1,29 @@
...
+    delete BackHandlerEmitterEvents[eventName];
+  }),
+  exitApp: jest.fn(),
+  fireEvent(eventName) {
+    BackHandlerEmitterEvents[eventName]();
+  }
+};
+
+// eslint-disable-next-line fp/no-proxy
+export const BackHandler = new Proxy(BaseBackHandler, {
```
Fausto95, line 21: I Kinda dislike this, will probably remove it tho :)

PR #539, opened 2020-05-28
Android : fix overflow bug on some correction cards

alanlanglois:
## Detailed purpose of the PR

On some cards, we could see the content of a card outside of the card itself (when the content is too long). 

| **Before**  | **After** |
| ------------- | ------------- |
|   |  |

### Testing strategy

- [x] I ensured my code is tested: lint / flow / unit / emulator


alanlanglois:
[Bug affichage cours "dépasser nos biais inconscients"](https://trello.com/c/RJytj7hw/1732-bug-affichage-cours-d%C3%A9passer-nos-biais-inconscients)

PR #542, opened 2020-06-10
Upgrade To React Native 0.62 🚀🎉

Fausto95:
This PR upgrades the current(0.59.9) React-Native version to the 0.62

This is a huge PR because we mostly upgraded all native non-native dependencies.
There are no visual breaking changes as well as technical.

For the record, Since there's an ongoing work to migrate from Flow to TypeScript(#490 ), I decided not to bump the current Flow's version to match the React-Native's one.

Previous attempts onto bumping RN's versions: #503 & #394.

## TODO (as a follow up PRs)
This are some of the work that needs to be done but aren't blocking.
- [ ] Rework entire camera permission workflow.
- [ ] Increase e2e performance & reduce running time.
- [ ] Rework some of the e2e logics
- [ ] Go through all of the animations in app, and turn them off(to stop) when component unmounts
- [ ] Fix some decisions taken to make this PR pass

### Testing strategy

- [x] I ensured my code is tested: 
- [x] I took care to mention breaking changes and extra libs if appropriate

PR #545, opened 2020-06-16
Fix react-native 0.62 android auto linking

Fausto95:
This PR lets React-Native handle the library linking itself(automatically).
This PR is a follow up of #542 

### Testing strategy


Fausto95:
@alanlanglois could you test that the app still launching and running correctly on your device, please?

alanlanglois:
> @alanlanglois could you test that the app still launching and running correctly on your device, please?

It introduces severals bugs:
- normal / fullscreen synchronization 
- subtitles broken

Those bugs are not visible on the 542.

PR #551, opened 2020-06-17
Dont shuffle the choices when e2e

Fausto95:
## Detailed purpose of the PR

> :wave: Write here :wink:

### Testing strategy


src/redux/actions/progressions/create-level-progression.js
```diff
@@ -19,7 +19,7 @@ export const createLevelProgression = (level: Level, engineVersion?: string) =>
     livesDisabled: level.infiniteLives
   };
 
-  if (__TEST__ || level.shuffleChoices === false) {
+  if (__TEST__ || __E2E__ || level.shuffleChoices === false) {
```
esa-coorp, line 22: il faut la même pour le microlearning

PR #553, opened 2020-06-19
Add GoogleTagManager [542 analytics follow up]

Fausto95:
#542 follow up :)

### Testing strategy


ios/Podfile
```diff
@@ -80,13 +80,8 @@ pod 'glog', :podspec => '../node_modules/react-native/third-party-podspecs/glog.
 pod 'Folly', :podspec => '../node_modules/react-native/third-party-podspecs/Folly.podspec'
 
 
-# # Monitoring
```
Fausto95, line 83: Handled by react-native's auto-linking

src/layer/data/logger.js
```diff
@@ -16,7 +16,7 @@ export const logError = async (error: Error) => {
   const {env: platform, name: brand} = _brand || {};
 
   await logDatadogError(error, {platform, brand, userId});
-  await crashlytics().recordError(0, error.stack || error.message);
+  await crashlytics().recordError(error);
```
Fausto95, line 19: newer version now requires an error as `recordError`'s argument

PR #554, opened 2020-06-19
Fix E2E Camera Permissions [#542 Follow Up]

Fausto95:
#542 follow up 


Fausto95:
I ran this on my machine and is working just as intended :)

PR #557, opened 2020-06-24
Adding notifyMe screen

alanlanglois:
## Detailed purpose of the PR

New "Notify Me" screen, 

### Testing strategy


src/components/notify-me.stories.tsx
```diff
@@ -0,0 +1,11 @@
+// @flow
+
```
Fausto95, line 2:
```suggestion
```

PR #568, opened 2020-07-07
Enhance permissions

Fausto95:
## Detailed purpose of the PR

> :wave: Write here :wink:

### Testing strategy


src/components/switch.stories.tsx
```diff
@@ -6,8 +6,8 @@ import Switch from './switch';
 
 const SwitchWithState = () => {
   const [isActive, setIsActive] = React.useState(true);
-  function handleOnPress() {
-    setIsActive((prev) => !prev);
+  async function handleOnPress() {
+    await setIsActive((prev) => !prev);
```
alanlanglois, line 10: No latency here? Could be frustrating if we don't have feedback right after we press

src/containers/app-session-listener.tsx
```diff
@@ -17,4 +16,5 @@ const AppSessionListener = ({incrementAppSession: incrementSession}: Props) => {
   return null;
```
alanlanglois, line 16: I'm note sure I understand which event ask for incrementation.

src/components/switch.stories.tsx
Fausto95: Yeah, it adds a bit of latency, not that frustrating.. but surely needs to be enhanced.

PR #575, opened 2020-07-22
Fix empty question screen after validating previous question

Fausto95:
Fixes an empty question screen after a previous question validation 

### Testing strategy


Fausto95:
Functionally tested it, works on my device

PR #577, opened 2020-07-22
Finish course notification AppState. 

alanlanglois:
## Detailed purpose of the PR
Adding listener for App focus:
 - when app is in foreground we cancel all notificaitons
 - when it goes inactive or background we setup all the needed notification for finish course

### Testing strategy


src/services/hero.ts
```diff
@@ -6,7 +6,7 @@ import type {
 import type {DataLayer} from '../layer/data';
 import {getAggregationsByContent} from '../layer/data/progressions';
 
-const isStepGTE3 = (aggregation: HeroRecommendation): boolean => aggregation.nbSlides >= 3;
+const isStepGTE3 = (aggregation: HeroRecommendation): boolean => aggregation.nbSlides >= 1;
```
Fausto95, line 9:
```suggestion
const isStepGTE3 = (aggregation: HeroRecommendation): boolean => aggregation.nbSlides >= 3;
```

Fausto95: I changed this, my mistake...

PR #579, opened 2020-07-23
Finish Course Notification[follow up]

Fausto95:
This PR handles wordings, schedule notification only when user want to receive them and also handles the navigation when user opens a notification

### Testing strategy


src/redux/actions/notifications/scheduled-notifications.ts
```diff
@@ -26,25 +28,44 @@ export type Action =
       };
     };
 
-const scheduleNotificationOnDevice = (title: string, message: string, date: Date) => {
+const getNotificationWording = () => {
+  const {finishCourseWordings} = translations;
+  return finishCourseWordings[Math.floor(Math.random() * finishCourseWordings.length)];
```
Fausto95, line 33: perhaps we need something more random here.
cc: @alanlanglois

alanlanglois: more random than math.random() ? haha
We could use an index and pick wording in a certain order, so if we have 5 translations we take the first, then the second, then the third etc and start again with the first... this way we would be sure to never have the same wording two times in a row. 
Or we could store the last id and ask a new random until it's different from the one we had before.

src/app.tsx
```diff
@@ -31,20 +35,33 @@ const reduxDevTools: ReduxDevTools | void =
...
+  },
 
-    // notif
-    // (required) Called when a remote is received or opened, or local notification is opened
-    notification.finish(PushNotificationIOS.FetchResult.NoData);
+  onNotification: function (notification: {
+    data: {id: string; content?: string};
+    finish: (arg: unknown) => void;
+  }) {
+    const {data} = notification;
```
Fausto95, line 54: I will try/catch this on a follow up PR

PR #590, opened 2020-08-13
Remove backslash from notification description

Fausto95:
Remove backslash's

### Testing strategy


src/redux/actions/notifications/scheduled-notifications.ts
```diff
@@ -45,7 +45,7 @@ const scheduleNotificationOnDevice = (
   if (!userName || !content.title) return;
   const notification = {
     title: title.replace('{{givenName}}', userName),
-    body: description.replace('{{contentName}}', content.title),
+    body: description.replace('\\', '').replace('{{contentName}}', content.title),
```
Fausto95, line 48:
```suggestion
    body: description.replace(/\\/g, '').replace('{{contentName}}', content.title),
```

PR #597, opened 2020-09-03
Exclude external content

alanlanglois:
## Detailed purpose of the PR
The mobile app only support course and chapter, before this PR users could see external content. We added a filter both side (mooc / app) to prevent external content to be loaded in the app. 

### Testing strategy

- [x] I ensured my code is tested: manual emulator android / test-unit. 


src/__fixtures__/cards.ts
```diff
@@ -212,7 +212,83 @@ export const createChapterCard = ({
...
+    completion,
+    title,
+    isAdaptive = false,
+    isNew = false,
+    isFavorite = false,
+    isDone = false,
+    status,
+    stars = 0,
+    authors = [createCardAuthor({})],
+    accessible = true,
```
Fausto95, line 229:
```suggestion
    accessible = true,
    type = 'scorm'
```

src/__fixtures__/cards.ts
```diff
@@ -212,7 +212,83 @@ export const createChapterCard = ({
...
+      {
+        ref: 'them_EJEGNqKn4',
+        label: 'Flawless French',
+      },
+    ],
+    groupsHidden: 'All courses, Flawless French',
+    course: null,
+    authors,
+    authorsListHidden: 'A good guy',
+    type: 'ext',
```
Fausto95, line 270:
```suggestion
    type,
```

Fausto95 (approved):
We will need to improve the createExternalContentCard function's return data to match the real one used in the web app.
I'm validating this one so we can fix the bug

alanlanglois:
> We will need to improve the createExternalContentCard function's return data to match the real one used in the web app.
> I'm validating this one so we can fix the bug

yep, I have no idea on how I can find those data. 
Is there specs or doc somewhere?

Fausto95:
Fixed by #602

PR #598, opened 2020-09-07
Fix completion for adaptive courses with more than a level

Fausto95:
Some courses that are adaptives happen to have more than a level, in this scenario, we actually want to show the completion based on them rather than alternating between 0 & 100%.

Capture d’écran 2020-09-07 à 11 19 50

No breaking changes for the other content's completions 🙂

### Testing strategy


alanlanglois:
You might change the test then?

alanlanglois (approved):

esa-coorp (approved):

PR #601, opened 2020-09-07
Notif suggestion (settings switch)

alanlanglois:
## Detailed purpose of the PR

Adding a new switch in settings for recommandation notifications.

### Testing strategy


src/redux/reducers/notifications/suggestion.ts
```diff
@@ -0,0 +1,30 @@
...
+export type State = {
+  type: NotificationType;
+  label: string;
+  isActive: boolean;
+};
+
+const initialState: State = {
+  type: NOTIFICATION_TYPE.SUGGESTION,
+  label: 'This course might be for you', //TODO add translations
+  isActive: Platform.OS === 'android',
```
Fausto95, line 17:
```suggestion
  isActive: true,
```

src/screens/settings.tsx
```diff
@@ -25,11 +26,15 @@ const styles = StyleSheet.create({
 export interface ConnectedStateProps {
   canReceiveNotifications: boolean;
   currentNotificationsPermission: PermissionStatus;
-  notificationsSettings: Record<'finishCourse', Pick<NotificationsState, 'finishCourse'>>;
+  notificationsSettings: Record<
```
Fausto95, line 29:
```ts
notificationsSettings: {
    'finish-course': Pick<NotificationsState, 'finishCourse'>,
    suggestion: Pick<NotificationsState, 'suggestion'>
}
```

src/screens/settings.tsx
```diff
@@ -65,6 +72,10 @@ const SettingsScreen = ({
           _toggleFinishCourseNotification();
           break;
         }
+        case NOTIFICATION_TYPE.SUGGESTION: {
+          _toggleSuggestionNotification();
+          break;
+        }
```
Fausto95, line 78: I think we could have a single toggle function that takes in a type and does the actual toggle.

Fausto95 (approved):

Fausto95 (changes requested):
See comments

src/redux/reducers/notifications/suggestion.ts
Fausto95: Don't take this comment is consideration,
But we should somewhere set this to true if the iOS user has already granted notifications permissions to us

src/screens/settings.tsx
alanlanglois: Yep, you're right it might be easier to maintain.

src/screens/settings.tsx
```diff
@@ -43,10 +38,20 @@ interface Props
     ConnectedDispatchProps,
     WithPermissionsProps {}
 
+export const transformNotificationSettings = (settings: NotificationsSettingsState) => {
```
Fausto95, line 41: For some reason doing this transformation screws up the order of the elements, hence the sorting

src/redux/actions/notifications/scheduled-notifications.test.ts
```diff
@@ -25,10 +25,12 @@ const createStore = (isActive: boolean, scheduledNotifications?: ScheduledNotifi
       },
     },
     notifications: {
-      finishCourse: {
-        type: 'finish-course',
-        label: 'Finish course',
-        isActive: isActive,
+      settings: {
+        'finish-course': {
+          type: 'finish-course',
```
alanlanglois, line 30: remove that

PR #605, opened 2020-09-09
Fixing headers on Android

alanlanglois:
## Detailed purpose of the PR

Fixing header position and size. It seems the RNNavigation update messed up on Android Headers :)

### Testing strategy


src/modules/status-bar.ts
```diff
@@ -1,11 +1,19 @@
...
     return 0;
   }
-
   return _getStatusBarHeight();
 };
+
+export const getHeaderHeight = (hedaerHeight = HEADER_HEIGHT): number => {
+  if (__STORYBOOK__) {
+    return hedaerHeight;
+  }
```
Fausto95, line 16: actually no need for this case

Fausto95 (approved):

PR #609, opened 2020-09-15
Refactor notifications settings status

alanlanglois:
## Detailed purpose of the PR

[iOS] Fixing the case when user already have the app installed with  access granted for notifications to set suggestion switch "on" by default.

### Testing strategy

- [x] I ensured my code is tested: ✅ test unit ✅ emulator / simulator  


Fausto95:
Could you constatify the status(just like NOTIFICATION_TYPE)?
Also, shouldn’t we set the suggestion setting to true on home screen did mount?

alanlanglois:
> Could you constatify the status(just like NOTIFICATION_TYPE)?
> Also, shouldn’t we set the suggestion setting to true on home screen did mount?

I'm on it :)

src/components/switch.tsx
```diff
@@ -38,11 +39,12 @@ const styles = StyleSheet.create({
 
 interface Props {
   onPress: () => Promise<void>;
-  isActive: boolean;
+  status: NotificationSettingStatus;
```
alanlanglois, line 42: should stay is active IMO, as switch could be used in other case not involving notifications

Fausto95: Agreed

Fausto95: So we need to change from `status` to `isActive` here: https://github.com/CoorpAcademy/mobile/blob/31b6944099b2d71d247d0ae8f2f6324c48278ed5/src/screens/settings.tsx#L41

alanlanglois: Are you sure?
I just changed this in the component:
`<Switch isActive={item.status === NOTIFICATION_SETTINGS_STATUS.ACTIVATED}
`
So we can keep a String argument for analytics

src/const.ts
```diff
@@ -169,6 +170,15 @@ export const RESOURCE_MIME_TYPE: Record<
   JPEG: 'image/jpeg',
 };
 
+export const NOTIFICATION_SETTINGS_STATUS: Record<
+  'IDDLE' | 'ACTIVATED' | 'DESACTIVATED',
+  NotificationSettingStatus
+> = {
```
Fausto95, line 176: IDLE

src/const.ts
```diff
@@ -169,6 +170,15 @@ export const RESOURCE_MIME_TYPE: Record<
...
+  IDDLE: 'idle',
+  ACTIVATED: 'activated',
+  DESACTIVATED: 'deactivated',
```
Fausto95, line 179: DEACTIVATED

Fausto95 (changes requested):

src/const.ts
alanlanglois: Excuse ma French...

Fausto95 (approved):

PR #610, opened 2020-09-21
Suggestion notification [follow up]

Fausto95:
#### Highlights

- Adds `Authorize notifications` settings
- Adds the ability to send different content for each suggestion notification scheduled
- Adds a `state reconciler` for redux-persist
- Fixes expired dev token.

### Testing strategy


src/const.ts
```diff
@@ -142,6 +143,14 @@ export const NOTIFICATION_TYPE: Record<'FINISH_COURSE' | 'SUGGESTION', Notificat
   SUGGESTION: 'suggestion',
 };
 
+export const NOTIFICATION_SETTINGS_TYPE: Record<
+  'FINISH_COURSE' | 'SUGGESTION' | 'AUTHORIZE',
```
alanlanglois, line 147: is this All ? 
If yes maybe call it ALL ?

alanlanglois: nevermind

alanlanglois (approved):

alanlanglois (approved):

PR #617, opened 2020-10-16
Redux toolkit

Fausto95:
Adding Redux-Toolkit and therefore adopting a new feature-based folder structure for Redux store.
I started working on this just to test RTK to see if it reduces the boilerplate and the complexity of current Redux style, it sure does.

RTK introduces new APIs to help create the store, the reducers(slices for instance) and manage them with ease.
I changed this the folder structure to remove the top-level `actions` and `reducers` folders so it reflects more the store itself, and as I was digging into the Redux docs, I figured out that it matches their new guidelines. so we'll stick with it.

As of this PR, I just refactored the easiest store pieces so that way we can avoid having a huge PR which refactors all of it.

The new folder structure looks like this: 

```
---- redux/
       ---- feature/
               ----- index.ts
               ----- index.test.ts
       ---- feature2/
               ----- actions/
                         ---- index.ts
                         ---- index.test.ts
               ----- reducers/
                         ---- index.ts
                         ---- index.test.ts
```

Note that we can have `actions` and `reducers` folders inside a feature folder when the flatten approach's index file becomes too big.

RTK has this `createSlice` API for creating piece of states which seems really cool, but for now I'm just using `createReducer` as it seems more simple to put in place. But later(or for new piece of state) we will use the new API.


alanlanglois (approved):
gg

